PR #69, opened 2015-04-22
Update import_logs.py

RMastop:
added ibooks as download extension


mattab:
Thanks @RMastop for the suggestion

PR #71, opened 2015-05-05
Improve ability to debug w/ various added diagnostic & debug messages/improvements.

diosmosis:
Includes the following changes:

* Do not truncate responses received from server when printing error messages.
* Add response + HTTP error code to error message if HTTP error exception is caught. The error message will look like, '404 Not Found: ... piwik response ...' instead of 'Not Found'.
* Output redirects in debug logs for easier debugging. With --debug supplied, redirects sent by Piwik will appear in the log output.

@quba Can you confirm this will help you debug log importer issues?
@mattab Can you do a quick review?

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -1135,6 +1135,17 @@ class Piwik(object):
 
             self.code = code
 
+    class RedirectHandlerWithLogging(urllib2.HTTPRedirectHandler):
+        """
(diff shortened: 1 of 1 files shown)


mattab:
Looks good to me - @quba do you need more information to make Log Analytics more useful?

PR #79, opened 2015-06-24
Add option for date format in logs.

marcenuc:
Custom log format can be specified with --log-format-regex option.
But there's no way to also specify a format string for the date.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -563,8 +563,12 @@ class Configuration(object):
                  "Overrides --log-format-name." % (', '.join(available_regex_groups))
         )
         option_parser.add_option(
+            '--log-date-format', dest='log_date_format', default=None,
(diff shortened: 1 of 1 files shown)


mattab:
Hi @marcenuc 
Thanks for the pull request! Would you mind adding an automated test in https://github.com/piwik/piwik-log-analytics/blob/master/tests/tests.py ? it would help make sure we don't break this feature in the future. Once done we could merge it

gothicx:
This is useful. I had to hard code a date format inside the .py code.

marcenuc:
I will add automated tests, but now I am too busy. Sorry... obviously, I will be glad if anyone else steps in to do it before me ;-)

diosmosis:
Closed in new PR #92

Issue #80, opened 2015-06-29
--add-sites-new-hosts add sites multiple times


chrodriguez:
When using json format and run script with `--add-sites-new-hosts` option, each time the script is run, new sites are created. 

Looking inside the code, it seems that when `_get_site_id_from_hit_host` is called an API call to `SitesManager.getSitesIdFromSiteUrl` is done without prefixing with *protocol*

I fixed it in my branch, so I will create a PR with the fix

halfdan:
What does your log look like? I think most log formats contain the protocol (which makes the lookup work). You may want to change that PR to only prefix `http://` if `hit.host` doesn't already contain the protocol.

chrodriguez:
I followed the README where it explains how to configure nginx to log json directly. The *host* parameter, from nginx documentation is defined as:

```
$host
in this order of precedence: host name from the request line, or host name from the “Host” request header field, or the server name matching a request
```

A line from my nginx json log format looks like:

```json
{
  "ip": "66.249.64.157", 
  "host": "www.extensionistas.unlp.edu.ar", 
  "path": "/article/printPreview/id/111", 
  "status": "200", 
  "referrer": "-", 
  "user_agent": "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)", 
  "length": 2730, 
  "generation_time_milli": 0.043, 
  "date": "2015-06-29T03:08:13-03:00"
}
```  

When I wrote the patch, I though it would be usefull to check if protocol is setted, but as variable was named host, I assumed it doesn't contains protocol. 

I will change it and update my PR

halfdan:
@chrodriguez You might be right - the protocol is added just a few lines later: https://github.com/chrodriguez/piwik-log-analytics/blob/a2316b97473227bff31ec62611f488c74763fdda/import_logs.py#L1323

chrodriguez:
@halfdan I've just create a new PR. I think this one solves the problem

chrodriguez:
I've tested it and it's working. Do you think this patch can be merged?

mattab:
It's working in 2.14.0 - issue is that you used latest version of log analytics with older Piwik (usually this works, but not here). we do not support running newer log analytics with older piwik (often it works but we don't support it)

Issue #77, opened 2015-06-08
import_logs.py should look for https host if no http host is found


rhurling:
Hi,

I currently have the problem that I want to auto-import my log files which didn't work, because the import_logs.py `DynamicResolver` class only looks for 'http://' + host in Piwik, but all my Sites are https only and created in piwik accordingly.

The following change to the `_get_site_id_from_hit_host` function would allow the script to look for a https host if no http host is found.

```python
    def _get_site_id_from_hit_host(self, hit):
        main_url = 'http://' + hit.host
        res = piwik.call_api(
            'SitesManager.getSitesIdFromSiteUrl',
            url=main_url,
        )
        if res:
            return res

        main_url = 'https://' + hit.host
        return piwik.call_api(
            'SitesManager.getSitesIdFromSiteUrl',
            url=main_url,
        )
```

mattab:
Hi @rhurling,
Thanks for the report, good find

mattab:
Note that to have this issue actually fixed we will need this Pull request merged: https://github.com/piwik/piwik/issues/8070

it will be done in the next Piwik 2.14.0 beta version available next week

PR #81, opened 2015-06-29
Fix getSitesIdFromSiteUrl API call

chrodriguez:
Call uses hit without protocol and this search will always fail and
create a new site every time this script is called

This bug is reported as #80

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -1314,9 +1314,13 @@ class DynamicResolver(object):
             self._cache['sites'] = piwik.call_api('SitesManager.getAllSites')
 
     def _get_site_id_from_hit_host(self, hit):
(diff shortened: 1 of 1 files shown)


chrodriguez:
Update PR y checking if host includes protocol or not

mattab:
@chrodriguez Can you try with Piwik 2.14.0-RC2 ? we made some changes recently, see: #77 and https://github.com/piwik/piwik/pull/8072 

Do you have the same problem with 2.14.0-RC2?

chrodriguez:
You are right! It seems to be fixed in the new release. However, we are not planning to upgrade piwik yet, so I think I will be using my fork until our upgrade.

Close this PR if you want

Thanks

PR #87, opened 2015-08-14
Improve documentation for --regex-group-to-visit-cvar --regex-group-to-page-cvar

mattab:


mattab:
fixed in https://github.com/piwik/piwik-log-analytics/commit/2a6f29858e7a80b23ec1601a979976a5353c4914

PR #90, opened 2015-08-18
add --date-format command line option

abone28:
Some web servers allow to change date/time format in logs (at least apache does). Currently import_logs.py can parse only a few common date/time formats, so it is hard to import logs with customized timestamps.

This patch adds --date-format option to import_logs.py, which allow to directly import logs with any date/time format that could be parsed by Pyhton's strptime.


abone28:
Opps. Looks like duplicate for #79. Sorry. Closing. This one.

PR #94, opened 2015-08-25
Use bulk tracking indices to display the lines of logs that were considered invalid by the Piwik tracker.

diosmosis:
As title.

Tests are included w/ https://github.com/piwik/piwik/pull/8644.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -1700,6 +1700,8 @@ class Recorder(object):
                     # debug tracker output will always break the normal JSON output.
                     if not config.options.debug_tracker:
                         logging.info("tracker response:\n%s" % response)
+
+                    response = {}
                 
                 if ('invalid_indices' in response and isinstance(response['invalid_indices'], list) and
                     response['invalid_indices']):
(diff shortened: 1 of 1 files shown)


import_logs.py
```diff
@@ -1655,6 +1671,30 @@ def _record_hits(self, hits):
...
+                    response = json.loads(response)
+                except:
+                    logging.info("bulk tracking returned invalid JSON")
+
+                    # don't display the tracker response if we're debugging the tracker.
+                    # debug tracker output will always break the normal JSON output.
+                    if not config.options.debug_tracker:
+                        logging.info("tracker response:\n%s" % response)
+                
+                if ('invalid_indices' in response and isinstance(response['invalid_indices'], list) and
```
tsteur: Would it trigger an exception if response is eg `null` because no json is returned? Not so familiar with Python anymore. Been a few years since I last used it.

diosmosis: It should, but I think response will be an empty string at worst. It could be invalid json too, which would create some odd-ness. Will fix.

PR #99, opened 2015-08-31
Make sure to always apply site_id if specified via option, even when --replay-tracking is used

### LINKED ISSUE #97: --idsite option seems to be ignored when used with --reply-tracking
It looks like it's set initially here https://github.com/piwik/piwik-log-analytics/blob/master/import_logs.py#L1591
and then overwritten here https://github.com/piwik/piwik-log-analytics/blob/master/import_logs.py#L1600

Example command:

```
/home/vagrant/www/piwik/misc/log-analytics/import_logs.py --url=http://apache.piwik/ --replay-tracking '/home/vagrant/www/testfiles/ecommerce_idsite_5.log' --idsite=5 2>&1
```

To test it one can use eg the ecommerce log file from Piwik repository: `plugins/VisitorGenerator/data/ecommerce.log`

It's weird as I thought I tested it initially and worked but now realize that it doesn't work. Also I'm not 100% if it's a bug or feature but does sound like a bug to me. Would need this feature for a custom plugin.

tsteur:
fixes #97 

We should merge https://github.com/piwik/piwik/pull/8677/files after merging this one

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -1599,6 +1599,10 @@ class Recorder(object):
 
         args.update(hit.args)
 
+        # option --idsite was used, we need to force using this siteId
+        if site_id and site_id != args['idsite']:
+            args['idsite'] = site_id
+
         if hit.is_download:
             args['download'] = args['url']
 
--- a/import_logs.py
+++ b/import_logs.py
@@ -1596,12 +1596,12 @@ class Recorder(object):
         if config.options.replay_tracking:
             # prevent request to be force recorded when option replay-tracking
             args['rec'] = '0'
-
+            
+        # idsite is already determined by resolver
+        if 'idsite' in hit.args:
+            del hit.args['idsite']
+            
         args.update(hit.args)
-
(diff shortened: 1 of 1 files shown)


import_logs.py
```diff
@@ -1599,6 +1599,10 @@ def _get_hit_args(self, hit):
 
         args.update(hit.args)
 
+        # option --idsite was used, we need to force using this siteId
+        if site_id and site_id != args['idsite']:
+            args['idsite'] = site_id
```
diosmosis: DynamicResolver will handle the case when replay tracking is used and when `--idsite` is passed, the StaticResolver instance is used (which returns just the idsite specified in the option), so I think instead of this code, it would be better to just remove `'idsite'` from `hit.args` before the `args.update(hit.args)`. Since idsite should be determined by the site resolver, and not `hit.args` (so it keeps the responsibility of handling --idsite to StaticResolver).

tsteur: done

PR #106, opened 2015-10-09
Add OVH log format

charlycoste:
OVH shared hosting are providing raw logs in a format which is a bit different from the common_vhost format. Actually, the host is not before, but after the IP address.

So I added a regex for this specific case, for my own needs. If you think it's interesting, just feel free to merge it...

And thank you for your contribution to libre software!

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -397,7 +397,8 @@ class AmazonCloudFrontFormat(W3cExtendedFormat):
         else:
             return super(AmazonCloudFrontFormat, self).get(key)
 
-_HOST_PREFIX = '(?P<host>[\w\-\.]*)(?::\d+)?\s+'
+_HOST_PREFIX = '(?P<host>([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,})(?::\d+)?\s+'
+
 _COMMON_LOG_FORMAT = (
     '(?P<ip>\S+)\s+\S+\s+(?P<userid>\S+)\s+\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s+'
     '"\S+\s+(?P<path>.*?)\s+\S+"\s+(?P<status>\S+)\s+(?P<length>\S+)'
@@ -419,6 +420,12 @@ _ELB_LOG_FORMAT = (
     '"\S+\s+\w+:\/\/(?P<host>[\w\-\.]*):\d+(?P<path>\/\S*)\s+[^"]+"\s+"(?P<user_agent>[^"]+)"\s+\S+\s+\S+'
 )
 
+_OVH_FORMAT = (
+    '(?P<ip>\S+)\s+' + _HOST_PREFIX + '(?P<userid>\S+)\s+\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s+'
+    '"\S+\s+(?P<path>.*?)\s+\S+"\s+(?P<status>\S+)\s+(?P<length>\S+)'
+    '\s+"(?P<referrer>.*?)"\s+"(?P<user_agent>.*?)"'
+)
+
 FORMATS = {
     'common': RegexFormat('common', _COMMON_LOG_FORMAT),
     'common_vhost': RegexFormat('common_vhost', _HOST_PREFIX + _COMMON_LOG_FORMAT),
@@ -432,6 +439,7 @@ FORMATS = {
     'icecast2': RegexFormat('icecast2', _ICECAST2_LOG_FORMAT),
     'elb': RegexFormat('elb', _ELB_LOG_FORMAT, '%Y-%m-%dT%H:%M:%S'),
     'nginx_json': JsonFormat('nginx_json'),
+    'ovh': RegexFormat('ovh', _OVH_FORMAT)
 }
 
 ##
--- a/tests/logs/ovh.log
+++ b/tests/logs/ovh.log
@@ -1,0 +1,1 @@
+1.2.3.4 www.example.com theuser [10/Feb/2012:16:42:07 -0500] "GET / HTTP/1.1" 301 368 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/535.11 (KHTML, like Gecko) Chrome/17.0.963.56 Safari/535.11"
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -387,6 +387,9 @@ def check_match_groups(format_name, groups):
     check_function = symbols['check_' + format_name + '_groups']
     return check_function(groups)
 
+def check_ovh_groups(groups):
+    check_common_complete_groups(groups)
+
 # parsing tests
 def test_format_parsing():
     # test format regex parses correctly


charlycoste:
Outch! Build failed... I'll check it out.

tsteur:
@charlycoste did you look into the build?

tsteur:
OK thx for the update. It would be really nice to have support for OVH as it is a big provider :)

tsteur:
Would it be maybe possible to otherwise change the test in some way to make it easier? How is the log format for OVH and icecast2 different? Is the only difference some spaces?

mattab:
Hello @charlycoste - Would you be able to help on this one? would be great

charlycoste:
Sorry I was a bit busy but I restarted to work on it last week. I'm not fluent in Python so give me just few more days and I will tell you if I'm definitly stuck.

charlycoste:
OK, sorry for the delay but I progressed a bit with this problem and I can summerize it.
The fact is that many formats are using the _common log format_ as a base which is like this : 

    127.0.0.1 - theuser [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326

Whereas the OVH format begins with something like this : 

    127.0.0.1 www.example.com theuser [10/Oct/2000:13:55:36 -0700] "GET / HTTP/1.1" 200 2326

Actually, there is a conflict between "-" and "www.example.com". In the common log format, "-" could be a "user identifier", the identifier used to identify the client making the HTTP request, which is not semanticaly the same as the user id. (see  [http://publib.boulder.ibm.com/tividd/td/ITWSA/ITWSA_info45/en_US/HTML/guide/c-logs.html#common](http://publib.boulder.ibm.com/tividd/td/ITWSA/ITWSA_info45/en_US/HTML/guide/c-logs.html#common))

In the code, this user identifier is not matched as a group `\S+` for the common log whereas my ovh regex matches it as the host `(?P<host>[\w\-\.]*)(?::\d+)?`

So, every regex that doesn't use the _HOST_PREFIX will match less groups than the OVH one.

charlycoste:
It seems that I could solve the problem by making _HOST_PREFIX a bit more strict at matching host names.

charlycoste:
@tsteur ? @mattab ? Did you have a look ?

G2v:
Waiting for this change to be committed, --log-format-regex `(?P<ip>\S+)\s+(?P<host>\S+)\s+(?P<userid>\S+)\s+\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s+"\S+\s+(?P<path>.*?)\s+\S+"\s+(?P<status>\S+)\s+(?P<length>\S+)\s+"(?P<referrer>.*?)"\s+"(?P<user_agent>.*?)"` matches OVH shared hosting logs

charlycoste:
up

mattab:
Thanks for the PR @charlycoste :+1: 

We will take a look in coming days!

sgiehl:
Actually I don't think we can fix that issue by 100%. The base problem is as mentioned, that the common log format has a user-identifier at the position where OVH has the host, everything else is the same.
Sure, changing the host pattern fixes the issue for that (test) case, but anyone could have a hostname like `dev` or `local`, which wouldn't match the regex. On the other side it isn't restricted to have a common log containing user-identifiers like `jan.in`, which would match the new host pattern.

Imho we should add the new log format, but without any changes to the host pattern and skip that format in auto detection. That would at least make it possible to force the log format using `--log-format-name=ovh` (as it won't be auto detected)
@mattab @tsteur any opinion?

<details>
<summary>I'd adjust this PR like this:</summary>

```diff
diff --git a/import_logs.py b/import_logs.py
index 403fc42..6479cce 100755
--- a/import_logs.py
+++ b/import_logs.py
@@ -399,7 +399,7 @@ class AmazonCloudFrontFormat(W3cExtendedFormat):
         else:
             return super(AmazonCloudFrontFormat, self).get(key)

-_HOST_PREFIX = '(?P<host>([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,})(?::\d+)?\s+'
+_HOST_PREFIX = '(?P<host>[\w\-\.]*)(?::\d+)?\s+'
 _COMMON_LOG_FORMAT = (
     '(?P<ip>\S+)\s+\S+\s+(?P<userid>\S+)\s+\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s+'
     '"\S+\s+(?P<path>.*?)\s+\S+"\s+(?P<status>\S+)\s+(?P<length>\S+)'
@@ -413,7 +413,7 @@ _S3_LOG_FORMAT = (
     '\S+\s+\S+\s+\S+\s+"(?P<referrer>.*?)"\s+"(?P<user_agent>.*?)"'
 )
 _ICECAST2_LOG_FORMAT = ( _NCSA_EXTENDED_LOG_FORMAT +
-    '\s+(?P<session_time>\S+)'
+    '\s+(?P<session_time>[0-9]+)'
 )
 _ELB_LOG_FORMAT = (
     '(?P<date>[0-9-]+T[0-9:]+)\.\S+\s+\S+\s+(?P<ip>\S+):\d+\s+\S+:\d+\s+\S+\s+(?P<generation_time_secs>\S+)\s+\S+\s+'
@@ -2011,6 +2011,9 @@ class Parser(object):
         for name, candidate_format in FORMATS.iteritems():
             logging.debug("Check format %s", name)

+            if name == 'ovh':
+                continue
+
             match = None
             try:
                 if isinstance(lineOrFile, basestring):

diff --git a/tests/tests.py b/tests/tests.py
index 2e1b64a..05d6246 100644
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -113,7 +113,7 @@ def test_format_detection():

     for format_name in import_logs.FORMATS.iterkeys():
         # w3c extended tested by iis and netscaler log files; amazon cloudfront tested later
-        if format_name == 'w3c_extended' or format_name == 'amazon_cloudfront':
+        if format_name == 'w3c_extended' or format_name == 'amazon_cloudfront' or format_name == 'ovh':
             continue

         f = functools.partial(_test, format_name)
@@ -657,6 +657,49 @@ def test_amazon_cloudfront_web_parsing():

     assert len(hits) == 1

+def test_ovh_parsing():
+    """test parsing of ovh logs (which needs to be forced, as it's not autodetected)"""
+
+    file_ = 'logs/ovh.log'
+
+    # have to override previous globals override for this test
+    import_logs.config.options.custom_w3c_fields = {}
+    Recorder.recorders = []
+    import_logs.parser = import_logs.Parser()
+    import_logs.config.format = import_logs.FORMATS['ovh']
+    import_logs.config.options.log_hostname = None
+    import_logs.config.options.enable_http_redirects = True
+    import_logs.config.options.enable_http_errors = True
+    import_logs.config.options.replay_tracking = False
+    import_logs.config.options.w3c_time_taken_in_millisecs = False
+    import_logs.parser.parse(file_)
+
+    hits = [hit.__dict__ for hit in Recorder.recorders]
+
+    assert hits[0]['status'] == u'301'
+    assert hits[0]['userid'] == u'theuser'
+    assert hits[0]['is_error'] == False
+    assert hits[0]['extension'] == u'/'
+    assert hits[0]['is_download'] == False
+    assert hits[0]['referrer'] == u''
+    assert hits[0]['args'] == {'uid': u'theuser'}
+    assert hits[0]['generation_time_milli'] == 0
+    assert hits[0]['host'] == 'www.example.com'
+    assert hits[0]['filename'] == 'logs/ovh.log'
+    assert hits[0]['is_redirect'] == True
+    assert hits[0]['date'] == datetime.datetime(2012, 2, 10, 21, 42, 07)
+    assert hits[0]['lineno'] == 0
+    assert hits[0]['ip'] == u'1.2.3.4'
+    assert hits[0]['query_string'] == ''
+    assert hits[0]['path'] == u'/'
+    assert hits[0]['is_robot'] == False
+    assert hits[0]['full_path'] == u'/'
+    assert hits[0]['user_agent'] == u'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/535.11 (KHTML, like Gecko) Chrome/17.0.963.56 Safari/535.11'
+
+    assert len(hits) == 1
+
+    import_logs.config.options.log_hostname = 'foo'
+
 def test_amazon_cloudfront_rtmp_parsing():
     """test parsing of amazon cloudfront rtmp logs (which use extended W3C log format w/ custom fields for event info)"""

```

mattab:
@sgiehl Good review and patch looks good. Let's add a note to the README file so it's easy to find?

sgiehl:
@charlycoste  could you apply the changes I had mentioned and maybe add a little note in the `Supported log formats`-section in README?

sgiehl:
@charlycoste I'll trackle that myself

PR #110, opened 2015-10-17
Issue when generation_time_milli is 0

gregoire-astruc:
Whenever generation_time_milli is 0 ( or more accurately, `0.0000`), the json returned will be interpreted as a *`string`* and not a *`float`*.

Forcing the cast fixes the issue.

Happened to me while serving static html pages.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -151,7 +151,7 @@ class JsonFormat(BaseFormat):
     def get(self, key):
         # Some ugly patchs ...
         if key == 'generation_time_milli':
-            self.json[key] =  int(self.json[key] * 1000)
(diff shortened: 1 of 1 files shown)


gregoire-astruc:
Okay, some more disclaimer here: I misread the README with all its quotes and accidentally put `"generation_time_milli": "$request_time"` in my configuration (notice the extra quotes).

So python was happily duplicating the string a thousand times. My bad on this one.

However this fix allows to handle both quoted and unquoted values, so use your best judgement :)

mattab:
Hi @gregoire-astruc 

Thanks for the PR

PR #116, opened 2015-12-10
Add tests for ipv6 parsing and make sure w3c extended format parses ipv6 addresses correctly.

### LINKED ISSUE #4: Make import_logs.py IPv6 compatible
Hi,

we are currently investigating if piwik logfile analytics can replace awstats log file analytics. 
Doing so we came across the problem that piwik seems not to be able to parse log files with IPv6 addresses. Removing every IPv6 log line from file and the import works fine.

The following import command is used 

```
python /var/www/piwik.ibumedia.de/misc/log-analytics/import_logs.py --url=http://piwik.ibumedia.de --idsite=77 --recorders=4 --enable-http-errors --enable-http-redirects --enable-static --enable-bots --recorder-max-payload-size=400 --dry-run 20130728.log
```

The log format is an iis 7 w3c format with all fields selected.
Lines like this will fail on import 

```
2013-07-28 02:34:09 W3SVC01 HOST123 1234:1234:12:123::1 GET / - 80 - 2a01:4f8:0:a101::6:1 HTTP/1.1 Hetzner+System+Monitoring - - www.example.com 200 0 0 3446 96 0
```

best regards 
mr.moe
Keywords: logfile ipv6 import

Migrated from piwik/piwik#4062

diosmosis:
As title.

New python tests will make sure every format will parse ipv6 ip addresses correctly.

Fixes #4

CC @mattab


mattab:
good that we could locate and fix this bug :+1:

Issue #111, opened 2015-11-07
Option to use stdin readline


falstaff84:
When using import_logs.py with Apache 2.4's CustomLog pipe option on a Ubuntu 12.04, I noticed that visits trickle in a bit delayed (on my machine 33 entries were necessary until the first visit got recorded). What is also interesting when shutting down the webserver, the remaining log entries do get lost! The python buffer options (-u) did not help to solve this issue. It would be nice to have an option to use the unbuffered sys.stdin.readline() when using CustomLog.

```
   -u     Force stdin, stdout and stderr to  be  totally  unbuffered.   On
          systems  where  it matters, also put stdin, stdout and stderr in
          binary mode.  Note that there is internal  buffering  in  xread-
          lines(),  readlines()  and  file-object  iterators ("for line in
          sys.stdin") which is not influenced by  this  option.   To  work
          around  this, you will want to use "sys.stdin.readline()" inside
          a "while 1:" loop.
```

antoninblanc:
The patch https://issues.piwik.org/3757 works for me.

mattab:
Hi @antoninblanc it would be useful if you can create a pull request on this repository with the working patch, test it, and then we will be happy to merge it for everyone to enjoy :+1:

mattab:
Hi @falstaff84 can you please confirm whether the patch in https://github.com/piwik/piwik-log-analytics/issues/118 fixes the issue for you too?

mattab:
Assuming https://github.com/piwik/piwik-log-analytics/pull/118 fixed this issue. If not, please reopen @antoninblanc @falstaff84

PR #118, opened 2015-12-18
Avoid Python stdin buffering

antoninblanc:
Pull request from patch https://issues.piwik.org/3757.

ref #111


mattab:
Thanks for the Pull request @antoninblanc

PR #120, opened 2015-12-28
Convert hex escapes in malformed JSON to unicode escapes to workaround nginx bug.

### LINKED ISSUE #16: import_logs.py doesn't support Unicode characters with json format
I'm working with iOS apps and sometimes their names have special characters and when this apps make requests to my server, the import_log.py script fails to parse this lines. This only happens when using nginx_json format.

Here's an example:

```
{"ip": "1.1.1.1","host": "blabla.com","path": "/api/test.xml","status": "200","referrer": "-","user_agent": "F\xFAtbol/1.0 (iPhone; iOS 7.1; Scale/2.00)","length": 267,"generation_time_milli": 0.009,"date": "2014-04-16T14:56:24-03:00"}
```

Check that user_agent has a special character and that the script fails when parsing this lines.

Thanks

Migrated from piwik/piwik#5013

diosmosis:
As title. Fix is simple.

Fixes #16

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -142,6 +142,11 @@ class JsonFormat(BaseFormat):
 
     def match(self, line):
         try:
+            # nginx outputs malformed JSON w/ hex escapes when confronted w/ non-UTF input. we have to
+            # workaround this by converting hex escapes in strings to unicode escapes. the conversion is naive,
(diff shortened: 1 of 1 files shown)


mattab:
@diosmosis Nice catch! and +1 for adding the tests in https://github.com/piwik/piwik/issues/9453

PR #127, opened 2016-02-06
Adds support for shoutcast

### LINKED ISSUE #114: How to import Shoutcast log files into Piwik for analysis
Hi,
I'm trying to extract information from Shoutcast streaming server logs but unfortunately struggling to set it up properly. I've seen some examples to extract Icecast log files which is a bit likewise. But I can't get it to work with Shoutcast. Hopefully anybody can help me out with this one;-) I'm trying to extract client details of listeners from the radio station (e.g. IP, streaming client software, etcetera).
I've added two log files the server generates, don't know which of the two is most suitable for import. Both seem to generate similar information. But I think that sc_serv.txt is the most readable log.

So hopefully anybody has the perfect recipe to import data into Piwik:-)

[sc_serv.txt](https://github.com/piwik/piwik-log-analytics/files/55085/sc_serv.txt)
[sc_w3c.txt](https://github.com/piwik/piwik-log-analytics/files/55086/sc_w3c.txt)

sgiehl:
changes proposed by @entropywaves 

fixes #114

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -401,7 +401,7 @@ def test_format_parsing():
 
     for format_name in import_logs.FORMATS.iterkeys():
         # w3c extended tested by IIS and netscaler logs; amazon cloudfront tested individually
-        if format_name == 'w3c_extended' or format_name == 'amazon_cloudfront':
+        if format_name == 'w3c_extended' or format_name == 'amazon_cloudfront' or format_name == 'shoutcast':
             continue
 
         f = functools.partial(_test, format_name, 'logs/' + format_name + '.log')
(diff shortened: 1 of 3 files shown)


sgiehl:
I've added a urldecode to the useragent and added some tests

tests/tests.py
```diff
@@ -534,6 +534,45 @@ def test_netscaler_parsing():
...
+    assert hits[0]['is_redirect'] == False
+    assert hits[0]['date'] == datetime.datetime(2015, 12, 7, 10, 37, 5)
+    assert hits[0]['lineno'] == 3
+    assert hits[0]['ip'] == u'1.2.3.4'
+    assert hits[0]['query_string'] == u'title=UKR%20Nights'
+    assert hits[0]['path'] == u'/stream'
+    assert hits[0]['is_robot'] == False
+    assert hits[0]['full_path'] == u'/stream?title=UKR%20Nights'
+    assert hits[0]['user_agent'] == u'NSPlayer/10.0.0.3702 WMFSDK/10.0'
+    assert hits[0]['length'] == 65580
```
mattab: @sgiehl could you add a test for the parsing of bandwith (value `524640`) which can be set as `&bw_bytes=` (useful when using plugin http://plugins.piwik.org/Bandwidth )?

sgiehl: If I see that correctly `bw_bytes` is set to the length (https://github.com/piwik/piwik-log-analytics/blob/shoutcast/import_logs.py#L1698-L1699)
I'm bit confused about that. The length of a request isn't the same as the bandwidth, or?!

mattab: The length of a request indicates the size of the returned response. Effectively this should be what we call 'bandwidth', which is the number of bytes of this response.

sgiehl: So should I set the `length` to `avgbandwidth` instead in this case? Or set the `bw_bytes` directly and change the mentioned code to set it only if not already set?

mattab: I tried searching the meaning of `sc-bytes` and `avgbandwidth` fields in the log format and found: https://www.microsoft.com/windows/windowsmedia/howto/articles/LoggingModel.aspx

* `sc-bytes` Total number of bytes the server sent to the client. The value does not include any overhead that is added by the network stack. However, protocols such as MMS, RTSP, and HTTP may introduce some overhead. Therefore, the same content streamed by using different protocols may result in different values. This field contains a hyphen (-) in propagated cache/proxy logs and in multicast log files.
* `avgbandwidth`: 	Average bandwidth (in bits per second) at which the client was connected to the server. The value is calculated across the entire duration of the connection.

so I suppose we set the `length` to `sc-bytes`, which will effectively set it as `bw_bytes` in Tracking API request?

sgiehl: Should be exacttly what is done atm, so I guess there is nothing to change.

mattab: oh nice, I missed it

entropywaves:
Is the code now incorporated into release 2.16.1 ?

Issue #138, opened 2016-05-11
Piwik & self signed certificates : urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed. Add new --accept-invalid-ssl-certificate


Daryes:
Hello,
This happens when using a self signed ssl certificate on a https hosted Piwik. 
It doesn't happen when falling back to http or downgrading python, something already referenced in #42 .
Of course, it's not possible to import any logs with this

> 2016-05-11 19:10:58,171: [INFO] Error when connecting to Piwik: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)>
> 2016-05-11 19:10:58,171: [INFO] Max number of attempts reached, server is unreachable!
> Traceback (most recent call last):
>   File "piwik/misc/log-analytics/import_logs.py", line 2338, in <module>
>     resolver = config.get_resolver()
>   File "piwik/misc/log-analytics/import_logs.py", line 972, in get_resolver
>     return StaticResolver(self.options.site_id)
>   File "piwik/misc/log-analytics/import_logs.py", line 1383, in **init**
>     'SitesManager.getSiteFromId', idSite=self.site_id
>   File "piwik/misc/log-analytics/import_logs.py", line 1366, in call_api
>     return cls._call_wrapper(cls._call_api, None, None, method, **kwargs)
>   File "piwik/misc/log-analytics/import_logs.py", line 1353, in _call_wrapper
>     raise Piwik.Error(message, code)
> __main__.Error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)

A bug I didn't experienced before upgrading the system and thus, doing the same with python jumping to v2.7.11. As expected, since Python had a change in v2.7.9+, making mandatory the certificate validation, which wasn't the case before. 
Redhat compiled some workarounds [on this page](https://access.redhat.com/articles/2039753), most of them working on a global level.

I've attached a dedicated workaround patch for the log-analytics I've done after searching around a lot, but I'm no python dev, it would be helpfull to add the whole mechanic for a --no-ssl-validity-check or similar parameter

[SSL-workaround.txt](https://github.com/piwik/piwik-log-analytics/files/260305/SSL-workaround.txt)

llsnndlldkm:
Today I received the same error message when I installed a Comodo EV-SSL SGC certificate:

[...]

PR #148, opened 2016-08-07
Do not validate HTTPS certificates

### LINKED ISSUE #42: When importing logs, if --url= is set to HTTP url and Piwik is forced to use SSL then importing logs fails
When [importing logs in Piwik](http://piwik.org/docs/log-analytics-tool-how-to/) and when Piwik is configured to use [SSL by default](http://piwik.org/faq/how-to/faq_91/), if you specify `--url=http://piwik.example.org` then Piwik will fail with this error:

```
2014-11-20 11:40:10,643: [DEBUG] Error when connecting to Piwik: <urlopen error Piwik returned an invalid response: <!DOCTYPE html>
```

The solution is to set the `--url` parameter to `https://....` when importing logs.

To fix this issue, we could maybe detect the URL redirect from HTTP to HTTPS and make Log importer detect that HTTPS should be used.

Migrated from piwik/piwik#6699

ntd:
Disclaimer: I am **not** a python programmer. This patch is the result of a single-day research that solved a problem I had (my google powers are clearly weak: I found issues #42 and  #138 after I solved the issue in my way).

When the HTTPS scheme is detected, do not perform certificate validation as forcibly required by recent Python changes.

Closes #42 and #138.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -1244,6 +1244,21 @@ class Piwik(object):
 
             return urllib2.HTTPRedirectHandler.redirect_request(self, req, fp, code, msg, hdrs, newurl)
 
+    class HTTPSHandlerWithoutValidation(urllib2.HTTPSHandler):
+        """
+        An HTTPS handler that does not validate its certificate, needed
+        when using self-signed certificates.
+        """
+
+        def __init__(self):
+            import ssl
+
+            context = ssl.create_default_context()
+            context.check_hostname = False
+            context.verify_mode = ssl.CERT_NONE
+
+            urllib2.HTTPSHandler.__init__(self, context = context)
+
     @staticmethod
     def _call(path, args, headers=None, url=None, data=None):
         """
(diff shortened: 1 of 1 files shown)


mattab:
Hi @ntd 

* this does not fix: https://github.com/piwik/piwik-log-analytics/issues/42  which is a different issue
* By default, we should always validate SSL certificates. so we must introduce a new parameter `--accept-invalid-ssl-certificate` where this behavior of ignoring SSL certificate validation would be enabled.

ntd:
> this does not fix: #42 which is a different issue

Hi @mattab,

well, the problem solved by this patch has been discussed there. Specifically you said to @shellshocker a pull request was welcome, so here it is. I think (not sure!) this is equivalent to the workaround proposed there in [comment 3](https://github.com/piwik/piwik-log-analytics/issues/42#issuecomment-207506965).

> By default, we should always validate SSL certificates. so we must introduce a new parameter --accept-invalid-ssl-certificate where this behavior of ignoring SSL certificate validation would be enabled.

Although I agree this would be the best option, I am not skilled enough to provide an implementation, sorry. Fill free to close this PR.

mattab:
Closing as we can't merge this

PR #152, opened 2016-09-16
fix nginx syslog link

cweiske:


mattab:
Thanks for that @cweiske

PR #153, opened 2016-09-27
Mention default nginx access log format

cweiske:


mattab:
Thanks, very appreciated @cweiske

PR #154, opened 2016-09-30
Mention correct nginx access log format

cweiske:
nginx 1.10.0's default access log format can be parsed with `ncsa_extended`. `common_complete` fails, and `common` does not contain referer and user agent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ To improve performance,
 
 ### Example Nginx Virtual Host Log Format
 
-nginx's default access log is parsed with the `--log-format-name=common_complete` option.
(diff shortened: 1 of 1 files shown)


mattab:
I hadn't double checked your previous PR https://github.com/piwik/piwik-log-analytics/pull/153/files nor did I double check this, please confirm this time it is confirmed correct on your end? Thanks

cweiske:
mattab, it's running with the mentioned format on the live server now.

The original format did only yield parsing errors, my first PR did then give me data. Only when I had the settings on the live server I noticed that referer and user agent were not getting parsed.

So all is fine now.

mattab:
Thanks for confirming and your prs. We hope you'll keep contributing :-)

PR #155, opened 2016-10-10
Support for custom bots

goetzk:
Because some sites have different, local bots this change allows those bots to
be passed in via options to the script rather than needing to modify the script
directly.

example usage:

```
python /srv/piwik-log-analytics/import_logs.py --url=https://example.piwik.pro/ --idsite=5 --enable-bots /var/log/nginx/*-access.log --custom-bot='_bot_' --custom-bot='Discourse.org'
```

This should suppress messages from Pingdom (which use Pingdom_bot_version) and
Discourse Server Monitoring.

This change is intended to compliment http://piwik.org/faq/how-to/faq_17483/

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -597,6 +597,10 @@ class Configuration(object):
             help="Track bots. All bot visits will have a Custom Variable set with name='Bot' and value='$Bot_user_agent_here$'"
         )
         option_parser.add_option(
+            '--custom-bot', dest='custom_bots', action='append', default=[],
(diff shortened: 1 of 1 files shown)


paskal:
If you want this imported, you'll need to rebase your commit and most likely get any votes on this issue (in my opinion, adding features just "because we can" is not a good approach). @mattab might have other opinion, but because it was not merged I think he'll agree.

goetzk:
After trailing Piwik we went with a different solution, so I'm not invested in seeing this change merged. 

thanks,
kk

PR #156, opened 2016-10-10
Add pingdom to EXCLUDED_USER_AGENTS list

goetzk:
Since pingdom is a popular service I thought it might be worth having on the
bot list. I won't be offended if the suggestion is to use _bot_ instead as it
would handle a wider variety of services.


mattab:
:+1:

PR #159, opened 2016-11-27
Fix nginx acess_log directive

paskal:
Tested with nginx/1.11.5


paskal:
Ubuntu comes with rsyslog, and I've added information about it's setup. I've tried to install syslog-ng an set ip up according to the article, but no success.

mattab:
Thanks @paskal for the suggestion :+1:

PR #168, opened 2017-03-19
Fix NGINX related documentation

easybe:
When logging multiple virtual hosts we need to use the
`common_complete` log format it is defined as
`host + ncsa_extended`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ The script will import all standard web server log files, and some files with no
  * log files with and without the virtual host will be imported
 
 In general, many fields are left optional to make the log importer very flexible.
- 
+
 ## Get involved
 
 We're looking for contributors! Feel free to submit Pull requests on Github.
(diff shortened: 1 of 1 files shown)


mattab:
Hello @easybe 

We used to have this in the documentation but @cweiske changed our doc in https://github.com/piwik/piwik-log-analytics/pull/154 6 months ago.

Please @cweiske can you discuss with @easybe and advise us what is the correct documentation or how we can clarify things better?

cweiske:
I guess both need to be listed, with a which to use when using multiple hosts and a single one.

easybe:
If you read the paragraph "Example Nginx Virtual Host Log Format" carefully you should see that there must be a copy & paste mistake. The first line tells you what format name to use with NGINX's default logging settings. But the last line should tell you which format to use in a virtual host setup. Christian Weiske noticed that `--log-format-name=common_complete` doesn't work with NGNIX's default settings. This makes sense because in the source code we see that `common_complete` is defined as `_HOST_PREFIX + _NCSA_EXTENDED_LOG_FORMAT`. He shouldn't have changed the second occurrence of `--log-format-name=common_complete` though.

cweiske:
@easybe you are right. The second should be reverted.

mattab:
Thanks for confirming guys, and for keeping our docs accurate :+1:

PR #170, opened 2017-04-18
Fixes a Bug with cvar "HTTP-code"

Legion2:
now the custom variable only gets added if it should


mattab:
Hi @Legion2 
Actually we want to set the HTTP-code for all pages.

PR #174, opened 2017-05-08
Mention tests/run_tests.sh script

fwolfst:
Probably every modern python-dev knows how that nosetests stuff works, but mentioning of `run_tests.sh` in README would have saved me some seconds.


mattab:
Thanks for the PR @fwolfst :+1:

PR #177, opened 2017-06-20
Improve documentation for --regex-group-to-visit-cvar --regex-group-to-page-cvar

mattab:
seems I forgot to merge this change 2 years ago


mattab:
No, it was actually merged already

PR #178, opened 2017-06-26
Allow wildcards for importing logs

geographika:
Related to https://github.com/piwik/piwik-log-analytics/issues/104 - the docs say this is possible, but I can't see from the code how this is implemented.


mattab:
@geographika I thought it is implemented by:
```

        for excluded_path in config.options.excluded_paths:
            if fnmatch.fnmatch(hit.path, excluded_path):
return False
```

in https://github.com/piwik/piwik-log-analytics/blob/master/import_logs.py#L1990-L1992 or was i maybe mistaken?

geographika:
Thanks for the quick reply! Apologies for the first incorrect pull request (I left a print statement in). 

That only seems to be checking the `--excluded_paths` option, whereas the docs seem to imply you can do the following to import all .log files in a folder (which didn't work for me prior to the changes in this pull request):

`python import_logs.py *.log`

The files to process are gathered by:

`self.options, self.filenames = option_parser.parse_args(sys.argv[1:])`

And then looped through later on with:

```
        for filename in config.filenames:
            parser.parse(filename)
```

mattab:
I see - it was probably working already here because linux bash will expand the *log and write the actual file names. 

Would it be possible to use the `fnmatch` instead of `glob` for consistency with the other wildcards?

geographika:
That would explain it. Yes I'm running this on a Windows server. 

I'd prefer `glob` as it has the advantage of being able to get files in subdirectories, whereas this would all need to be coded with `fnmatch` and `os.lisdir` and some recursion (`glob` uses `fnmatch` internally). 

The  check for a * wildcard is a bit hacky though. Could any log filenames on linux contain a *? This isn't possible on Windows, but is technically possible on Unix - https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

import_logs.py
```diff
@@ -824,7 +825,10 @@ def _parse_args(self, option_parser):
         if not self.filenames:
             print(option_parser.format_help())
             sys.exit(1)
-
+				
+        if len(self.filenames) == 1 and '*' in self.filenames[0]:
```
mattab: Why writing `len(self.filenames) == 1`, could we apply the glob when there are are more than 1 pattern specified?

PR #182, opened 2017-06-30
Improves automatic format detection (regex improvements)

sgiehl:
For most fields we currently use`\S+` as regex. This is very unspecific and matches anything.
For some fields we know the specific format it should be in, so it's better to exactly check for them:

- for IPs: `[\w*.:-]+`
- for HTTP status: `\d+`
- for session_time: `[0-9-]+`   (as it might be an integer or `-`)

This should improve the automatic format detection.


import_logs.py
```diff
@@ -401,23 +401,23 @@ def get(self, key):
...
     '\S+\s+\S+\s+\S+\s+"(?P<referrer>.*?)"\s+"(?P<user_agent>.*?)"'
 )
 _ICECAST2_LOG_FORMAT = ( _NCSA_EXTENDED_LOG_FORMAT +
-    '\s+(?P<session_time>\S+)'
+    '\s+(?P<session_time>[0-9-]+)'
 )
 _ELB_LOG_FORMAT = (
-    '(?P<date>[0-9-]+T[0-9:]+)\.\S+\s+\S+\s+(?P<ip>\S+):\d+\s+\S+:\d+\s+\S+\s+(?P<generation_time_secs>\S+)\s+\S+\s+'
-    '(?P<status>\S+)\s+\S+\s+\S+\s+(?P<length>\S+)\s+'
+    '(?P<date>[0-9-]+T[0-9:]+)\.\S+\s+\S+\s+(?P<ip>[\w*.:-]+):\d+\s+\S+:\d+\s+\S+\s+(?P<generation_time_secs>\S+)\s+\S+\s+'
```
mneudert: Is the asterisk included to support anonymized log files? Just stripping it made no difference for the tests.

sgiehl: Actually didn't think about that before. Just copied it from https://github.com/piwik/piwik-log-analytics/blob/master/import_logs.py#L229
But guess it would make sense for anonymized logs...

PR #196, opened 2017-11-10
Allow filtering logs by host & datetime

diosmosis:
Refs https://github.com/piwik/piwik-log-analytics/issues/194
Refs https://github.com/piwik/piwik-log-analytics/issues/193

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -799,7 +799,20 @@ class Configuration(object):
             '--request-timeout', dest='request_timeout', default=DEFAULT_SOCKET_TIMEOUT, type='int',
             help="The maximum number of seconds to wait before terminating an HTTP request to Piwik."
         )
+        option_parser.add_option(
+            '--include-host', action='callback', type='string', callback=functools.partial(self._add_to_array, 'include_host'), # TODO
+            help="Only import logs from the specified host(s)."
+        )
+        option_parser.add_option(
+            '--exclude-host', action='callback', type='string', callback=functools.partial(self._add_to_array, 'exclude_host'),
+            help="Only import logs that are not from the specified host(s)."
+        )
         return option_parser
+
+    def _add_to_array(self, option_attr_name, option, opt_str, value, parser):
+        if not hasattr(parser.values, option_attr_name):
+            setattr(parser.values, option_attr_name, [])
+        getattr(parser.values, option_attr_name).append(value)
 
     def _set_option_map(self, option_attr_name, option, opt_str, value, parser):
         """
@@ -2098,6 +2111,17 @@ class Parser(object):
         logging.debug('Format %s is the best match', format.name)
         return format
 
+    def is_filtered(self, hit):
+        if config.options.exclude_host and len(config.options.exclude_host) > 0:
+            if hit.host in config.options.exclude_host:
+                return True
+
+        if config.options.include_host and len(config.options.include_host) > 0:
+            if hit.host not in config.options.include_host:
+                return True
+
+        return False
+
     def parse(self, filename):
         """
         Parse the specified filename and insert hits in the queue.
@@ -2346,6 +2370,10 @@ class Parser(object):
                 except UnicodeDecodeError:
                     invalid_line(line, 'invalid encoding')
                     continue
+
+            if self.is_filtered(hit):
+                invalid_line(line, 'filtered out')
+                continue
 
             hits.append(hit)
 
--- a/import_logs.py
+++ b/import_logs.py
@@ -809,16 +809,28 @@ class Configuration(object):
         )
         option_parser.add_option(
             '--exclude-older-than', action='callback', type='string', default=None, callback=functools.partial(self._set_date, 'exclude_older_than'),
-            help="Ignore logs older than the specified date. Exclusive. Date format must be YYYY-MM-DD hh:mm:ss."
+            help="Ignore logs older than the specified date. Exclusive. Date format must be YYYY-MM-DD hh:mm:ss +/-0000. The timezone offset is required."
         )
         option_parser.add_option(
             '--exclude-newer-than', action='callback', type='string', default=None, callback=functools.partial(self._set_date, 'exclude_newer_than'),
-            help="Ignore logs newer than the specified date. Exclusive. Date format must be YYYY-MM-DD hh:mm:ss."
+            help="Ignore logs newer than the specified date. Exclusive. Date format must be YYYY-MM-DD hh:mm:ss +/-0000. The timezone offset is required."
         )
         return option_parser
 
     def _set_date(self, option_attr_name, option, opt_str, value, parser):
-        date = datetime.datetime.strptime(value, '%Y-%m-%d %H:%M:%S')
+        try:
+            (date_str, timezone) = value.rsplit(' ', 1)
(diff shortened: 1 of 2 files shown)


import_logs.py
```diff
@@ -2098,6 +2123,31 @@ def detect_format(file):
...
+                pass
+
+        if host:
+            if config.options.exclude_host and len(config.options.exclude_host) > 0 and host in config.options.exclude_host:
+                return True
+
+            if config.options.include_host and len(config.options.include_host) > 0 and host not in config.options.include_host:
+                return True
+
+        if config.options.exclude_older_than and hit.date < config.options.exclude_older_than:
```
tsteur: Is it possible that hit.date is not set as date assignment is wrapped in a try/catch? Not sure what should happen if hit.date is not set... I think as long as this comparison does not result in an error it should be fine... 

actually, I see further below it calls `hit.date.date()` so we can likely assume the date is already always set.

However, I think you want to use `>` instead of `<` and also swap it for newer comparison. I did eg `./import_logs.py --dry-run --url=http://apache.piwik --exclude-older-than="2017-11-10 00:00:23" ../../test4939.log` which should have removed no entry as all log files are from Nov 19th. However, it imported all. Then I did `./import_logs.py --dry-run --url=http://apache.piwik `--exclude-newer-than="2017-11-10 00:00:23" ../../test4939.log` and it imported none. This was when using `dry-run` feature and I can send you the log file if needed. The comparison looks correct so wondering if it might be different issue.

import_logs.py
```diff
@@ -2098,6 +2123,31 @@ def detect_format(file):
...
+        if hasattr(hit, 'host'):
+            host = hit.host
+        else:
+            try:
+                host = urlparse.urlparse(hit.path).hostname
+            except:
+                pass
+
+        if host:
+            if config.options.exclude_host and len(config.options.exclude_host) > 0 and host in config.options.exclude_host:
```
tsteur: This is currently an `in`. @mattab I wonder if we need an exact comparison here? or maybe different parameters one for equals and one for contains? In our case, a "contains" could match another host that we do not want to import.

tsteur:
@diosmosis 

Currently it shows filtered logs as "invalid". I wonder if it makes sense to group them together as "filtered logs"

```
    323 requests imported successfully
    0 requests were downloads
    697 requests ignored:
        2 HTTP errors
        0 HTTP redirects
        624 invalid log lines
        0 requests did not match any known site
        0 requests did not match any --hostname
        9 requests done by bots, search engines...
        62 requests to static resources (css, js, images, ico, ttf...)
        0 requests to file downloads did not match any --download-extension
```

In my case most of the requests were actually filtered.

Also I could not 100% verify yet, does this filter work with `--replay-tracking`?

import_logs.py
diosmosis: ‘exclude_host’ is a list.

tsteur: Sorry all good, when I tested it, I looked wrong at the result.

mattab:
yes it would be great to have 2 new lines below "X requests ignored" that say something like:
```
   Y requests hostnames did not match --include-host or did match --exclude-host 
   Z requests datetime did match --exclude-older-than or --exclude-newer-than
```

import_logs.py
tsteur: FYI: I debugged it and I think it works. The problem may be that `hit.date` as defined in log eg `[19/Nov/2017:00:00:01 +0100]` will be converted to UTC or something, while the excluded dates are not.

I reckon this is kind of expected but maybe also a little confusing? As a user someone will need to be careful. Maybe we could require to also specify timezone with the excluded dates or so? Any thoughts @diosmosis @mattab

diosmosis: Simplest solution is to say the times must be in UTC in the help output.

mattab: Suggestions:

* The excluded date should support the notation eg. `19/Nov/2017:00:00:01 +0100` or any other valid datetime such as `19/Nov/2017:00:00:01 NZST`  and convert it to UTC
* To clarify and avoid any user error I suggest we require user to input the timezone in the parameter. if timezone is missing, throw an error.

diosmosis: FYI parsing timezone info is not available in python's std lib: https://stackoverflow.com/questions/13182075/how-to-convert-a-timezone-aware-string-to-datetime-in-python-without-dateutil

Can assume an external library like `dateutil` is available (requiring users to have it installed), use an email date format like `Tue May 08 15:14:45 +0800 2012`, force the timezone value to be in a numeric format (so `NZST` will not be supported), or find a timezone parser and include it with this repo. What do you want me to do?

tsteur: We don't need timezone but similar to log format be able to specify `+/-XXXX` as in `19/Nov/2017:00:00:01 +0100` to match same date as specified in the logs. This way a user basically doesn't need to think and transform the `+0100` manually.

diosmosis: Then I'll require that a numeric timezone format is used.

tsteur: Awesome 👍

diosmosis:
> However, I think you want to use > instead of < and also swap it for newer comparison. I did eg ./import_logs.py --dry-run --url=http://apache.piwik --exclude-older-than="2017-11-10 00:00:23" ../../test4939.log which should have removed no entry as all log files are from Nov 19th. However, it imported all. Then I did ./import_logs.py --dry-run --url=http://apache.piwik--exclude-newer-than="2017-11-10 00:00:23" ../../test4939.logand it imported none.

@tsteur can you provide me w/ the log file that caused this error?

tsteur:
@diosmosis I believe this error was actually due to the fact that the date was expected to be in UTC and got confused with the invalid log lines output I think. By the looks it works I'd say 👍

diosmosis:
@tsteur I just pushed the timezone change, if it's still an issue w/ that log file, let me know.

diosmosis:
@tsteur @mattab Added a `filtered log lines` summary output line. Did not add one line per type of exclusion, if it's important let me know.

diosmosis:
Add a test for replay tracking in the piwik/piwik repo.

I noticed date filtering works, host filtering, however, won't if the host isn't found in the log file and if the host also isn't in the `url` query param to piwik.

tsteur:
looks good now @diosmosis 

> won't if the host isn't found in the log file and if the host also isn't in the url query param to piwik.

So it will work when it is found in the log file but not in the URL query right? Because we will need this feature for replay tracking and the host is in the log file itself.

tsteur:
Tested replay tracking and all seems to work well. Good to merge for me @mattab

diosmosis:
> So it will work when it is found in the log file but not in the URL query right? Because we will need this feature for replay tracking and the host is in the log file itself.

It should work in both cases (as long as the importer extracts the host correctly; if the format used doesn't parse out the host, then it won't work). I'll try and add a test for this in the other PR.

diosmosis:
There might be an issue if the host in the log file is different from the host in the `url` query parameter. @tsteur do you think the `url` query parameter host should be given higher priority than the log file host?

tsteur:
I would personally prefer the host in the log line, not in the piwik.php

diosmosis:
Ok, I'll leave it alone.

PR #199, opened 2017-12-11
Add --accept-invalid-ssl-certificate option.

### LINKED ISSUE #42: When importing logs, if --url= is set to HTTP url and Piwik is forced to use SSL then importing logs fails

smarsching:
This option allows a user to ignore certificate validation problems when connecting to the Piwik server. This is useful when the server uses a self-signed or otherwise invalid certificate, but obviously should only be used when the connection can be trusted.

This PR should resolve #42 and #138.

This PR implements the new --accept-invalid-ssl-certificate option in a slightly different way than suggested in #42: Instead of messing around with internal functions of the `ssl` module, it only uses public API by creating an SSL context using the default options, disabling validation for this context instance, and then passing this context to `urllib2` (or more precisely `HTTPSHandler`). This is also the way suggested by the Python documentation and - in contrast to the way suggested in #42 - should also work with Python 3.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -29,6 +29,7 @@ import os
 import os.path
 import Queue
 import re
+import ssl
 import sys
 import threading
 import time
@@ -815,6 +816,13 @@ class Configuration(object):
             '--exclude-newer-than', action='callback', type='string', default=None, callback=functools.partial(self._set_date, 'exclude_newer_than'),
             help="Ignore logs newer than the specified date. Exclusive. Date format must be YYYY-MM-DD hh:mm:ss +/-0000. The timezone offset is required."
         )
+
+        option_parser.add_option(
+            '--accept-invalid-ssl-certificate',
+            dest='accept_invalid_ssl_certificate', action='store_true',
+            default=False,
+            help="Do not verify the SSL / TLS certificate when contacting the Piwik server."
+        )
         return option_parser
 
     def _set_date(self, option_attr_name, option, opt_str, value, parser):
@@ -1358,7 +1366,16 @@ class Piwik(object):
             base64string = base64.encodestring('%s:%s' % (auth_user, auth_password)).replace('\n', '')
             request.add_header("Authorization", "Basic %s" % base64string)        
 
-        opener = urllib2.build_opener(Piwik.RedirectHandlerWithLogging())
+        # Use non-default SSL context if invalid certificates shall be
+        # accepted.
+        if config.options.accept_invalid_ssl_certificate:
+            ssl_context = ssl.create_default_context()
+            ssl_context.check_hostname = False
+            ssl_context.verify_mode = ssl.CERT_NONE
+        else:
+            ssl_context = None
+        opener = urllib2.build_opener(Piwik.RedirectHandlerWithLogging(),
+                                      urllib2.HTTPSHandler(context=ssl_context))
         response = opener.open(request, timeout = timeout)
         result = response.read()
         response.close()


mattab:
Hi @smarsching - thanks for the pull request! It works and fixes https://github.com/piwik/piwik-log-analytics/issues/138

mattab:
FYI https://github.com/piwik/piwik-log-analytics/issues/42 was not fixed, as it's a different issue (where force_ssl=1 is used, then importing logs specifying http:// will not work)

mattab:
@smarsching 
Unfortunately it seems this change actually broke the log importer in some cases. please see our build results here: https://travis-ci.org/piwik/piwik/jobs/315001887

```
There were 14 errors:

1) Piwik\Tests\System\ArchiveCronTest

Failed to setup fixture: log importer failed: 2017-12-11 20:55:26,783: [DEBUG] Accepted hostnames: all

2017-12-11 20:55:26,783: [DEBUG] Piwik Tracker API URL is: http://localhost/tests/PHPUnit/proxy/

2017-12-11 20:55:26,783: [DEBUG] Piwik Analytics API URL is: http://localhost/tests/PHPUnit/proxy/

2017-12-11 20:55:26,783: [DEBUG] No token-auth specified

2017-12-11 20:55:26,783: [DEBUG] No credentials specified, reading them from "/home/travis/build/piwik/piwik/config/config.ini.php"

2017-12-11 20:55:26,876: [DEBUG] Authentication token token_auth is: 3db1fa6860c3ef85e06ac06661aba0d1

2017-12-11 20:55:26,876: [DEBUG] Resolver: static

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 2527, in <module>

    resolver = config.get_resolver()

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1080, in get_resolver

    return StaticResolver(self.options.site_id)

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1522, in __init__

    'SitesManager.getSiteFromId', idSite=self.site_id

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1505, in call_api

    return cls._call_wrapper(cls._call_api, None, None, method, **kwargs)

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1455, in _call_wrapper

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1439, in _call_api

    res = Piwik._call('/', final_args, url=url)

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1390, in _call

    urllib2.HTTPSHandler(context=ssl_context))

TypeError: __init__() got an unexpected keyword argument 'context'

command used: python2.6 "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py" -ddd --url="http://localhost/tests/PHPUnit/proxy/" --idsite="1" --enable-testmode --recorders="1" --recorder-max-payload-size="1" "/home/travis/build/piwik/piwik/tests/resources/access-logs/fake_logs.log" 2>&1
```


---> Would be awesome if you could take a look and possibly issue a new pull request to fix this regression?

smarsching:
Thanks for the heads-up. My guess is that it is an issue with Python 2.6 (I only tested it with 2.7). I will check this and get back to you.

smarsching:
My suspicion was right, the `context` parameter of the `HTTPSHandler` constructor was first introduced in Python 2.7.9 (this was when they changed the default behavior so that certificates are validated). I will soon create a PR that adds a check whether we are running on Python >=2.7.9 and only passes the context in this case.

smarsching:
I just created PR #200, which should fix this problem.

PR #200, opened 2017-12-12
Make code compatible with Python <=2.7.8.

smarsching:
This fixes a regression introduced with PR #199.

The `context" parameter of the `urllib2.HTTPSHandler` constructor was only introduced in Python 2.7.9, so we cannot pass a specific SSL context in older Python versions. Luckily, we do not have to either because Python before 2.7.9 did not validate remote certificates anyway, so the behavior specified by the `--accept-invalid-ssl-certificate` option is actually the default when running
on these versions of Python.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -833,7 +833,7 @@ class Configuration(object):
             '--accept-invalid-ssl-certificate',
             dest='accept_invalid_ssl_certificate', action='store_true',
             default=False,
-            help="Do not verify the SSL / TLS certificate when contacting the Piwik server."
+            help="Do not verify the SSL / TLS certificate when contacting the Piwik server. This is the default when running on Python 2.7.8 or older."
         )
         return option_parser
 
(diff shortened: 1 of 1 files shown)


mattab:
@smarsching 
We're now getting another error: 

```
2017-12-12 22:35:32,555: [DEBUG] Using credentials: (login = superUserLogin, password = 1e56c228742c0189d261500852e27a02)

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 2528, in <module>

    config = Configuration()

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 990, in __init__

    self._parse_args(self._create_parser())

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 970, in _parse_args

    self.options.piwik_token_auth = self._get_token_auth()

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1009, in _get_token_auth

    _url=self.options.piwik_api_url,

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1508, in call_api

    return cls._call_wrapper(cls._call_api, None, None, method, **kwargs)

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1458, in _call_wrapper

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1442, in _call_api

    res = Piwik._call('/', final_args, url=url)

  File "/home/travis/build/piwik/piwik/misc/log-analytics/import_logs.py", line 1383, in _call

    if config.options.accept_invalid_ssl_certificate and \

NameError: global name 'config' is not defined
```

mattab:
Here is the build: https://travis-ci.org/piwik/piwik/jobs/315604395

Thanks in advance for your help on this

smarsching:
Hm, this is a tricky one: `_get_token_auth` in the `Configuration` class uses the `Piwik` object for making an HTTP request and with the new `--accept-invalid-ssl-certificate` option, we use the `config` object when making this HTTP request, so the HTTP request cannot be sent because the configuration is not available yet and the configuration cannot be initialized because the HTTP request cannot be sent.

The only idea that I have so far is splitting the initialization of the configuration into two parts: First initialize everything that does not depend on any component, so that the `config` object is available and then in a second pass initialize the remaining options.

From a first glance, it looks like the `piwik_token_auth` option is the only option where this problem appears, so it should be sufficient to move the initialization into a new method (maybe `init_token_auth` and calling this one after the `config = Configuration()` line like this:

```
…
piwik = Piwik()
config = Configuration()
config.init_token_auth()
stats = Statistics()
…
```

In fact, I am even thinking about moving the `piwik = Piwik()` line after the `config = Configuration` line to make it clear that the `piwik` object depends on the `config` object.

PR #201, opened 2017-12-13
Run _get_token_auth after config object is initialized.

smarsching:
This fixes the problem discussed in #200. The `_get_token_auth` method in the `Configuration` class may use the `piwik` object which depends on the `config` object being present. This means that the `_get_token_auth` method must not be called from the `Configuration` contructor, but only after the `config` object has been created.


mattab:
Thanks for the quick new PR :+1: will update the submodule in https://github.com/piwik/piwik/pull/12350 and wait for build results

PR #204, opened 2018-02-19
Update shebang to stick to Python version 2

J0WI:


sgiehl:
I think this would break #195

J0WI:
It doesn't break #195, but it will select Python 2 automatically, if installed. #195 is only thrown if you try to force Python 3.

dbrgn:
#242 is probably the better solution. Only 3 months left until Python 2 is officially unsupported: https://pythonclock.org/

sgiehl:
Closing this in favor of #242

PR #205, opened 2018-02-21
support Python3 (Proof of concept)

Findus23:
This is a quick test of a python3 compatible version.

Apart from converting the file with 2to3 only few manual changes were made, but it seems to be working in a simple use case and no tests are failing.

This version doesn't support python2. 
Either we could make this changes with a lot of `if`s or maybe keep both versions, deprecate the python2 one and keep the new one for eventual future refactoring.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: python
 
 python:
-  - "2.6"
-  - "2.7"
+  - 3.5
+  - 3.6
   # 3.0 is not supported
 
 sudo: false
--- a/import_logs.py
+++ b/import_logs.py
@@ -15,34 +15,34 @@
 
 import sys
 
-if sys.version_info[0] != 2:
+if sys.version_info[0] == 2:
     print('The log importer currently does not work with Python 3 (or higher)')
     print('Please use Python 2.6 or 2.7')
     sys.exit(1)
 
 import base64
 import bz2
-import ConfigParser
+import configparser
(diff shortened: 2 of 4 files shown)


Findus23:
The failing test seems to be because the order here jumps around in python3:
https://github.com/matomo-org/matomo-log-analytics/blob/4ace0ba5d879d19013327df2a77dff4e6b5a3275/tests/tests.py#L825

mattab:
Very useful.. :+1: 

> Either we could make this changes with a lot of ifs or maybe keep both versions, deprecate the python2 one and keep the new one for eventual future refactoring.

Moving on to Python 3 only makes sense. 

## some thoughts on the Python3 roll-out

* we could have the two scripts for a while and only build new features in the Python 3 version. (important fixes would go to both)
* it would be useful if the Python3 script failed with an easy to understand message, when python2 is used instead
* we would need to wait for Matomo 4.0.0 before making the switch and replace the current script by the Python 3 script for all users.
  * This would break by default all people who use the script as Python2 wouldn't run it and people explicitely use python2. Wondering if it's maybe a bad idea. If we do this, we need at least to make sure there is a clear error message, and clear communication in our Matomo 4 changelogs
  * even after Matomo 4 we could maybe keep the old Python2 script, at least for a while (maybe until Matomo 5 or so).

Findus23:
See #242 for the up to date branch

PR #220, opened 2018-07-22
Use --matomo-domain instead of --piwik-domain

sgiehl:
relates https://github.com/matomo-org/matomo/pull/13192


tsteur:
It should be fine to use it by now. It be only a problem if someone was using an older version of Matomo and a newer log analytics. So should be all fine to merge 👍

PR #222, opened 2018-09-27
Add haproxy log format

fdellwing:
See https://forum.matomo.org/t/regex-fuer-haproxy/29959 for some background (in german).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Build status (master branch) [![Build Status](https://travis-ci.org/matomo-org/m
 
 
 The script will import all standard web server log files, and some files with non-standard formats. The following log formats are supported:
- * all default log formats for: Nginx, Apache, IIS, Tomcat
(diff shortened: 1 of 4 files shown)


fdellwing:
It was not easy to get `nose` installed, but now it should be good to merge.

Findus23:
@fdellwing That's why it is planned to switch to python3 and also use a modern test system instead.
https://github.com/matomo-org/matomo-log-analytics/pull/205

fdellwing:
I don't know why the autodetection does not work, is it important that it works?

fdellwing:
Any progress on this?

JackSlateur:
The following regexp match on my logs (haproxy httplog):
`'(?P<ip>[\w*.]+).*\[(?P<date>.*)\].*\ (?P<status>\\b\d{3}\\b)\ (?P<length>\d+)\ -.*\"(?P<method>\S+)\ (?P<path>\S+).*'`

PR #228, opened 2018-10-25
Fixed Cloudfront user agent parsing and made test more realistic

shadow-light:
Matomo's previous Cloudfront test is outdated as doesn't use tabs as separators, is missing several fields, and most importantly doesn't double quote the values.

This pull request updates the test to be more realistic and fixes the parsing of the user agent. Without this Matomo does **not** detect browsers properly.


shadow-light:
See the following regarding double quoting (%2520 especially)

https://forums.aws.amazon.com/thread.jspa?messageID=484509

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html#LogFileFormat

import_logs.py
```diff
@@ -404,7 +404,7 @@ def get(self, key):
             return '200'
         elif key == 'user_agent':
             user_agent = super(AmazonCloudFrontFormat, self).get(key)
-            return urllib2.unquote(user_agent)
+            return urllib2.unquote(urllib2.unquote(user_agent))  # Value is double quoted!
```
sgiehl: According to AWS docs the useragent should only be quoted once. But guess it shouldn't hurt to unquote a second time, as it shouldn't change anything

shadow-light: That's what I thought, but from practice it was quoted twice for some reason.

Issue #144, opened 2016-07-12
Log Analytics could detect log lines that were already imported and skip them automatically


mattab:
[Log Analytics](http://piwik.org/log-analytics/) is a powerful tool of the Piwik platform, and used by thousands of people in many interesting use cases. It is quite powerful and relatively easy to use, and offers has many options and features. We like to make our tools as easy as possible to use... this issue is about making Log Analytics easier to use and even more flexible for people.

## Issue: the log data is not deduplicated

When you import logs in Piwik, Piwik will always import and track all the logs. When you import again the same log file in Piwik, it will be imported again by the Tracking API, and the data will end up being duplicated in the Piwik database. 

## Why this is not good enough

our users rightfully expect Piwik to be easy to use and do the right thing. Recently @Synchro reported this issue and did not expect Log Analytics to import the data again and again. See the description at: https://github.com/piwik/piwik/issues/10248#issuecomment-232085596

Over the years many users have reported experiencing this issue.

## Existing workaround

So far most people manage to use Log Analytics despite this limitation. The common workaround is to create one log file per hour, or one log file per day, and import each log file only once. Commonly, people write a script  which makes sure that each log file is imported only once. For example, the log files may be ingested into Piwik while/after they have been rotated.

## Solution

Ideally, we do not want people to worry whether they have imported a given log file, or even whether a log file was partially imported before and is re-imported again.  We want Piwik to automatically [deduplicate](https://en.wikipedia.org/wiki/Data_deduplication) the [tracking API](http://developer.piwik.org/api-reference/tracking-api) data.

so far I see two possible ways to fix this issue:

### 1. new Piwik Tracking API feature: request id deduplicator 

The [Tracking API](http://piwik.org/docs/tracking-api/) could introduce a new feature, to let tracking api users specify a `request ID` for the given request. Piwik would store the `request ID` for each request and use this `request ID` as a unique key. If any tracking API request for a given `date` with a given `request ID` has already been tracked/imported in this `date`, then the request would be skipped. Each `request id` will be imported at most once for a given day. 

The Log Analytics tool will then simply create, for each log file's line that is parsed, a request ID and pass it to the tracking api request to let the Tracking API deduplicate the requests. Log analytics could create this request id as a hash of the log line or so. 

* Pros:  other Tracking API SDKs and clients will be able to use this feature to deduplicate the data.

### 2. Implement request ID deduplicator in Log Analytics only

[...]

PR #232, opened 2018-11-17
Add an option to only process log entries that haven't been processed before

mackuba:
I want to use Matomo with log analytics only. My Nginx logs are rotated every week, but I want my reports to be updated much earlier, e.g. every hour. If I just feed the same log file with already reported visits to the importer, I will have duplicated entries, so I need to either rotate logs every hour (very inconvenient) or somehow prevent logs from being imported twice. Based on what I could find, there is currently no easy way to do this.

This pull request solves this by tracking the latest visit timestamp found in an imported log file and then saving it to a file specified in a `--timestamp-file` option. On the next run this timestamp is loaded at startup and all visits before or on this timestamp are ignored (like `--exclude-older-than`, but inclusive, since the log with equal timestamp was already parsed).

This kind of solves https://github.com/matomo-org/matomo-log-analytics/issues/144.

I've put `initial_timestamp` (loaded from the file at the beginning) in the config and `latest_timestamp` (updated after every log record) in the stats. This can be moved elsewhere if it's not the best place.

I've also added some lines to the summary to print the status of the timestamp-based filtering, and included the older/newer than filtering too since it's related:

```
Logs import summary
-------------------

    85 requests imported successfully
    36 requests were downloads
    10627 requests ignored:
        73 HTTP errors
        2 HTTP redirects
        0 invalid log lines
        345 filtered log lines
        0 requests did not match any known site
        0 requests did not match any --hostname
        153 requests done by bots, search engines...
        10054 requests to static resources (css, js, images, ico, ttf...)
        0 requests to file downloads did not match any --download-extensions

    Processed logs since: 2018-11-06 19:05:12 +0000
    Saved last timestamp: 2018-11-07 11:59:42 +0000
```

I also tweaked the printing there to remove extra empty lines (more than 2 newlines are compacted into 2) - this was already a problem before, as the space between the 2nd and 3rd section was bigger than between 1/2 and 3/4 because of `%(sites_ignored)s`, but was made more visible with the date filtering section added.


cweiske:
Works fine here.
The only problem I see is that the timestamp file gets updated even when using `--dry-run`.

DevDavido:
This is definitely a useful enhancement. Would definitely love to see that in the Matomo Log Importer 👍 .

DevDavido:
This is definitely a useful enhancement. Would certainly love to see that in the Matomo Log Importer 👍 .

strager:
I resolved merge conflicts with 4.x-dev (commit 6f66f962fcc985e998c9c6f1bf4522067bc5c07a) here: https://github.com/strager/matomo-log-analytics/tree/timestamp

strager:
ping. I've been using my version of this patch for a while and I've been happy with it.

atom-box:
Help wanted!

PR #235, opened 2018-11-29
By default support replaying tracking requests to both piwik.php and matomo.php.

### LINKED ISSUE #224: Support replay of piwik.php and matomo.php
In https://github.com/matomo-org/matomo/pull/13596 / https://github.com/matomo-org/matomo/issues/12785 we add support for matomo.php as endpoint besides piwik.php. It will also be the default for new installations.

The log importer needs to be adjusted to support replaying those requests when using eg `--replay-tracking`. It isn't that trivial since we currently can only specify one `--replay-tracking-expected-tracker-file=piwik.php` file but people might use both `piwik.php` and `matomo.php`. By default, the log importer should check for both files.

As for the `_record_hits ` method, we maybe still need to use `piwik.php` as tracking endpoint until Matomo 4.0 as it could otherwise break the import for existing users. We could set it to `matomo.php` and hope that if it breaks, users don't automatically delete log files so they can fix the issue and replay again after making `matomo.php` accessible. But realistically, users might only notice after few weeks that it broke and by then the log files may have been deleted. It's tricky. Maybe we could record into `matomo.php` by default, and if there is eg a 404, or 403 error, then we track into `piwik.php`?

diosmosis:
Also add new `--tracker-endpoint-path` to allow tracking to custom tracking endpoint.

Fixes #224

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -691,9 +691,9 @@ class Configuration(object):
             help="Replay piwik.php requests found in custom logs (only piwik.php requests expected). \nSee https://matomo.org/faq/how-to/faq_17033/"
         )
         option_parser.add_option(
-            '--replay-tracking-expected-tracker-file', dest='replay_tracking_expected_tracker_file', default='piwik.php',
-            help="The expected suffix for tracking request paths. Only logs whose paths end with this will be imported. Defaults "
-            "to 'piwik.php' so only requests to the piwik.php file will be imported."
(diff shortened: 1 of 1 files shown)


tsteur:
Looks good. @diosmosis can't see the `--tracker-endpoint-path` in the diff?

diosmosis:
Forgot to push, added now.

PR #236, opened 2018-11-29
Fix timezone calculation for non-integer timezones

mackuba:
There are two places in the code which apply a timezone difference to a parsed date: when parsing command line arguments (`Configuration._set_date`) and when processing a hit (`Parser.parse`). They do that in a rather hacky way: the timezone is parsed as a number and then the difference is set to be `number / 100` hours.

This kind of works... except it assumes that timezones can only be at full integer intervals, and that's not true. There are several non-integer timezones, including in such important countries as India (+05:30) or Iran (+03:30). In that case, `"0530"` is interpreted as 5.3 hours ahead of UTC instead of 5.5...

For example, with such log line:

`1.2.3.4 - - [29/Nov/2018:15:30:00 +0530] "GET / HTTP/1.1" 200 6355 "-" "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:43.0) Gecko/20100101 Firefox/43.0" 0.014`

the 15:30 Indian Standard Time should be read as 10:00 UTC (11:00 CET). Instead, it's 11:12:

screenshot 2018-11-29 at 17 18 41

This PR makes a relatively small fix to the timezone calculating logic in that the parsed "number" (not really an actual number) e.g. `530` is divided into hours and minutes separately using integer division and modulo and passed to `datetime.timedelta` as separate parameters. With that change, the timestamp is parsed as 10:00 UTC (11:00 CET) correctly:

screenshot 2018-11-29 at 17 52 48

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -852,10 +852,8 @@ class Configuration(object):
         if not re.match('[-+][0-9]{4}', timezone):
             fatal_error("Invalid date value '%s': expected valid timzeone like +0100 or -1200, got '%s'" % (value, timezone))
 
-        timezone = float(timezone)
-
         date = datetime.datetime.strptime(date_str, '%Y-%m-%d %H:%M:%S')
-        date -= datetime.timedelta(hours=timezone/100)
+        date -= TimeHelper.timedelta_from_timezone(timezone)
 
         setattr(parser.values, option_attr_name, date)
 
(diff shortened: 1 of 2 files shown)


sgiehl:
The changes look good for me. But as I'm not working with python regularly maybe some else could have a look

PR #237, opened 2019-01-14
Apply config.options.seconds_to_add_to_date to _...ts tracker params …

diosmosis:
…so they stay in sync w/ cdt.


sgiehl:
Is it possible to add a simple test for that?

diosmosis:
@tsteur fyi, merging this. was for enterprisedemo only

PR #238, opened 2019-01-26
fixed robots.txt not being counted as static file

mackuba:
The `robots.txt` file is listed in the `STATIC_EXTENSIONS` list, however the entries from this list are only compared against `hit.extension` - which in case of `robots.txt` will be `txt`. So the result is that this file is not excluded from log entries and appears on the Downloads report.

This PR adds a second list of static files with full filenames (with just this one file at the moment) that's instead compared with the whole last path element of the URL.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -66,7 +66,11 @@ except ImportError:
 ##
 
 STATIC_EXTENSIONS = set((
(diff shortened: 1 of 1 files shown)


sgiehl:
Works as expected. I'll add a test for that after merging

PR #242, opened 2019-02-14
support Python3 

### LINKED ISSUE #3: import_logs.py needs upgrade to Python 3 or updated shebang line
Some distributions like Arch have `/usr/bin/python` point to `python3`, not Python 2.x. [Even more popular distributions are transitioning to do the same thing](https://wiki.ubuntu.com/Python/3). Currently, the `import-logs.py` script does not work under Python 3. As such, either it should be updated to support Python 3, either its shebang line should be changed to `#!/usr/bin/python2` instead of assuming that `/usr/bin/python` is Python 2.x.
Keywords: log-analytics, python

Migrated from piwik/piwik#3759

### LINKED ISSUE #243: python2's bz2-module does not support multi-stream files
The bz2 implementation in python 2.7 is unable to decompress multi-stream bzip-files. This is especially problematic because it does not actually output an error or a warning and just acts as if the file was parsed completely [0].

When using the debug flag you will see something like this:
```
…
2019-03-28 12:54:57,119: [DEBUG] Invalid line detected (line did not match): xxx.xxx.xxx.xxx - - [03/Feb/2019:08:42:34 +0100] "GET /resources/some.pdf HTTP/2.0" 200 811576 "-" "Mozilla/5.0 (iPhone; CPU iPhone OS 12_1_2 like Mac OS X)
…
```

After this, the parsing for the file ends. This happens because the content of the uncompressed log-file is not separated into streams be line (but by size), so you will most likely have the first and only stream the bz2-module actually reads end in the middle of a long line.
In the files I tried the first stream ended after exactly 900.000 bytes, so that could be an indication.

To check your bz2-file you might want to try running bzip2 on console (this is a multi-stream file):

```bash
~$ bzip2 -tvvv ../logs/access.log-20190203.bz2 
  ../logs/access.log-20190203.bz2: 
    [1: huff+mtf rt+rld {0x0af23f00, 0x0af23f00}]
    [2: huff+mtf rt+rld {0xb291b802, 0xb291b802}]
    combined CRCs: stored = 0xa775c602, computed = 0xa775c602
    [1: huff+mtf rt+rld {0x12f00a0d, 0x12f00a0d}]
    [2: huff+mtf rt+rld {0xcbad012a, 0xcbad012a}]
…
    combined CRCs: stored = 0xfcecf793, computed = 0xfcecf793
    [1: huff+mtf rt+rld {0x84243689, 0x84243689}]
    combined CRCs: stored = 0x84243689, computed = 0x84243689
    ok
```

**Possible solutions**
- check the file-pointer after reading the compressed logfile, if it's exactly 900.000 bytes, output a warning or an error message
- Use an external dependency like bz2file [1] to read the compressed data

[0] https://docs.python.org/2/library/bz2.html#bz2.BZ2File
[1] https://pypi.org/project/bz2file/

Findus23:
followup to #205 (done from scratch)
fixes #3 
fixes #243

Making the code python3-compatible is easy: 2to3 does most of the tedious things, some minor changes have to be applied and nosetests gets replaced with pytest
One could of course say ["all tests are passing, so this can be merged"](https://travis-ci.org/Findus23/piwik-log-analytics/builds/493393077), but I feel like this may get much more complex and could surface some hard-to-notice bugs:

For example somewhere in [this function](https://github.com/matomo-org/matomo-log-analytics/blob/master/import_logs.py#L1333-L1365) it is assumed that dictionaries are always returned in the same order. 
But unlike in PHP, python dictionaries are always unordered and therefore this code may print `1\n2`, but may also print `2\n1`.
```python
example = {1: "one", 2: "two"}
for number, name in example.items():
    print(number)
```
While this difference was mostly in theory for python2, it also happens in practice with python3 and therefore [this test](https://travis-ci.org/Findus23/piwik-log-analytics/jobs/493390486) randomly fails to correctly decode the query string in 3.5.

Funfact: Starting with 3.7 dicts are officially ordered in python.

I cought this issue randomly as it occurs in the test on travis, but I have no idea how many similar cases there are in the rest of the code.

As I have never used loganalytics and I honestly don't understand large portions of the code, this needs to be tested extensively (with all 69 options, as things may break in subtle ways as e.g. GET parameters tracked slightly incorrect) and ideally read by someone with far more python (and especially python2) experience. 

BTW: I targeted Python 3.5, 3.6 and 3.7 as these are the versions mostly used and are easy to support at the same time.


sjbronner:
Hi everybody. I am very happy that `python3` support is seeing progress. Thanks @Findus23 for taking that on. I would like to offer a suggestion to solving "shebang problem".

### Problem Definition

As it is, when a user has a mismatched version of python available under `/usr/bin/python`, python will complain loudly about syntax errors or other problems with the script.

### Solution

Most python developers that I have seen will always specify the python version in the shebang: `#!/usr/bin/python3` or `#!/usr/bin/python2`. The reasoning behind that is that python 3 and python 2 are considered to be different languages. While they are similar, and one grew out of the other, they are entirely incompatible and thus different languages. That leads to the notion that specifying just `#!/usr/bin/python` does not actually specify the language that the script should be interpreted as.

This then automatically leads to a relevant error message when the needed version of python is not available. And as far as I know, all distributions provide files named `/usr/bin/python2` and `/usr/bin/python3`. Both Ubuntu and Arch Linux do, anyway.

### Suggestion

My suggestion is to add the needed python version number to the shebang in both the old and the new versions of this script.

Findus23:
@sjbronner Thanks for your detailed comment.

Honestly I think that the problem you are describing isn't really one. 
> As it is, when a user has a mismatched version of python available under `/usr/bin/python`, python will complain loudly about syntax errors or other supposed issues with the script.

No, the python3 script is written in a way so that you can run in with python2 and will see [the proper error message](https://github.com/matomo-org/matomo-log-analytics/pull/242/files#diff-3e1701183de6c665909d9b66c2a1de57R20) (which of course needs to be worded better).

And as much as I am a fan of writing linux-scripts, your solution doesn't solve the issue you describe as:
- it doesn't stop anyone from running the python3 script with python2 (the shebang is only read by the system, not by python)
- it assumes python can be found in ``/usr/bin/python3` (there is nothing that forces this)
- a common workaround is to instead use `/usr/bin/env python3`, but this also assumes that env is installed (in this path)
- this also assumes that this is the python version the user wants to use (instead of a venv somewhere else)

Of course adding a shebang still won't hurt, but that would be a seperate change and not have anything to do with python2/python3

Findus23:
Just had an idea on how to avoid the weird dict-related bugs:
Since python 3.1 the [`OrderedDict`](https://docs.python.org/3.7/library/collections.html#collections.OrderedDict) data-type is supported. So we could simply everywhere where the logic depends on dicts being ordered, replace the dicts with ordereddicts.

dbrgn:
Thanks for your work, @Findus23!

Only a bit more than 3 months left until Python 2 support is dropped: https://pythonclock.org/ After that, there will be no more security updates for Python 2, which makes running it on a production server a (growing) security risk.

benabel:
> Python 2.7 will not be maintained past 2020. Originally, there was no official date. Recently, that date has been updated to January 1, 2020.
Do you want some help on this PR. Maybe we could add more environments test(ubuntu, mac, windows) using github actions instead of travis-ci.

import_logs.py
```diff
@@ -1474,16 +1442,15 @@ def _call(path, args, headers=None, url=None, data=None):
...
         else:
             https_handler_args = {}
-        opener = urllib2.build_opener(
-            Matomo.RedirectHandlerWithLogging(),
-            urllib2.HTTPSHandler(**https_handler_args))
+        opener = urllib.request.build_opener(
+            self.RedirectHandlerWithLogging(),
+            urllib.request.HTTPSHandler(**https_handler_args))
         response = opener.open(request, timeout = timeout)
         result = response.read()
```
cbay, line 1448: You need to .decode() the response, as otherwise it would be returned as bytes, which e.g. json.loads() cannot use.

cbay:
```
open_func = gzip.open
```

Due to https://bugs.python.org/issue13989, it won't work, you can change it to:

```
open_func = lambda filename, mode: io.TextIOWrapper(gzip.open(filename))
```

import_logs.py
sgiehl: @cbay  are you able to point out a test case where the current implementation would fail? Tests are currently running, so would be good to add a test if possible...

cbay: The current test suite only tests a small part of the import script. It doesn't actually try to import hits to a Matomo instance, which is where the issue lies.

Actually adding such tests would be a massive task, I'm afraid.

sgiehl: Oh. I see. Would that be covered by the tests done on Matomo side? It uses the log importer in some test fixtures to import log files: https://github.com/matomo-org/matomo/blob/3.x-dev/tests/PHPUnit/Fixtures/ManySitesImportedLogs.php#L110

cbay: Yes, I think so!

I've also tested the script myself today (with the 2 aforementioned changes), it seems to work fine, at least for my use case. It has successfully parsed/recorded several millions lines. Using Python 3.4.

sgiehl: awesome. Would you mind to add your changes as suggestions, so we can simply apply them?

import_logs.py
```diff
@@ -10,56 +10,46 @@
 #
 # For more info see: https://matomo.org/log-analytics/ and https://matomo.org/docs/log-analytics-tool-how-to/
 #
-# Requires Python 2.6 or 2.7
+# Requires Python 3.5, 3.6 or 3.7
```
cbay, line 13: Any reason lower versions are not supported? I have used Python 3.4, it seems to work fine.

import_logs.py
```diff
@@ -10,56 +10,46 @@
...
+from __future__ import print_function  # this is needed that python2 can run the script until the warning below
 
 import sys
 
-if sys.version_info[0] != 2:
-    print('The log importer currently does not work with Python 3 (or higher)')
-    print('Please use Python 2.6 or 2.7')
+if sys.version_info[0] != 3:
+    print('The log importer currently does not support Python 2 any more.')
+    print('Please use Python 3.5, 3.6 or 3.7')
```
cbay: 3.8 works as well?

import_logs.py
```diff
@@ -564,346 +567,320 @@ def _create_parser(self):
...
+        return date
 
     def _parse_args(self, option_parser):
         """
         Parse the command line args and create self.options and self.filenames.
         """
-        self.options, self.filenames = option_parser.parse_args(sys.argv[1:])
+        self.options = option_parser.parse_args()
+        self.filenames=self.options.file
+        print(self.options)
```
cbay: To be removed.

import_logs.py
```diff
@@ -564,346 +567,320 @@ def _create_parser(self):
...
-        getattr(parser.values, option_attr_name)[key] = value
+        return date
 
     def _parse_args(self, option_parser):
         """
         Parse the command line args and create self.options and self.filenames.
         """
-        self.options, self.filenames = option_parser.parse_args(sys.argv[1:])
+        self.options = option_parser.parse_args()
+        self.filenames=self.options.file
```
cbay: Missing spaces around the =

tests/test_main.py
```diff
@@ -363,7 +354,7 @@ def check_iis_groups(groups):
                                'status', 'length', 'host', 'userid', 'generation_time_milli',
                                '__win32_status', 'cookie', 'method']
 
-    for property_name in groups.keys():
+    for property_name in list(groups.keys()):
```
cbay:
```suggestion
    for property_name in groups.keys():
```

import_logs.py
```diff
@@ -1396,20 +1364,21 @@ def _convert_dict_to_array(d):
...
 
         def __init__(self, message, code = None):
-            super(Matomo.Error, self).__init__(message)
+            super(MatomoHttpBase.Error, self).__init__(message)
 
             self.code = code
 
-    class RedirectHandlerWithLogging(urllib2.HTTPRedirectHandler):
+
+class MatomoHttpUrllib(MatomoHttpBase):
```
cbay, line 1375: Why making 2 classes (`MatomoHttpBase` and `MatomoHttpUrllib`)? No other class inherits `MatomoHttpBase`.

import_logs.py
```diff
@@ -2367,12 +2331,6 @@ def filtered_line(line, reason):
             if not line: break
             lineno = lineno + 1
 
-            try:
-                line = line.decode(config.options.encoding)
-            except UnicodeDecodeError:
-                invalid_line(line, 'invalid encoding')
-                continue
-
```
cbay, line 2375: I'd rather keep that code and open the files in binary mode instead, which lets us keep the `encoding` option and the `invalid encoding` statistic.

import_logs.py
sgiehl:
```suggestion
        self.filenames = self.options.file
```

import_logs.py
sgiehl:
```suggestion
```

import_logs.py
sgiehl: Tests are running against 3.8-dev, so guess it should work

import_logs.py
sgiehl: Will add running tests for 3.4 on travis

.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: python
-
+dist: xenial
 python:
-  - "2.6"
-  - "2.7"
-  # 3.0 is not supported
+  - "3.5"
```
sgiehl, line 4:
```suggestion
  - "3.4"
  - "3.5"
```

import_logs.py
dbrgn: Python 3.4 has reached end-of-life last year: https://www.python.org/downloads/release/python-3410/ I would advise against trying to actively support it when doing a new port.

import_logs.py
cbay: I didn't know you could do suggestions on code reviews! That's nice.

import_logs.py
sgiehl: Good point

Findus23: That was also the reason I left it out from the beginning. As it is end-of-life and far older than what I'm familiar with, I can't promise that anything works there.

import_logs.py
Findus23: When I wrote this, 3.8 wasn't released yet, but I don't know any reason why it wouldn't work. (3.8 is stable now, so we can test against the stable version.

tests/test_main.py
Findus23: I can't think of a reason why the list() would be required.

import_logs.py
Findus23: Because I planned to create a second (far simpler) Http class that uses requests for sending HTTP requests.

cbay: I don't think there's much to gain having 2 similar implementations. In any case, inheriting just for the `Error` class seems pointless, that class could be global anyway.

import_logs.py
sgiehl:
```suggestion
    print('Please use Python 3.5, 3.6, 3.7 or 3.8')
```

.travis.yml
```diff
@@ -1,9 +1,10 @@
...
+  - "3.6"
+  - "3.7"
+  - "3.8-dev"
```
sgiehl:
```suggestion
  - "3.8"
```

sgiehl:
guess we can close this one, as all changes are already included in 4.x-dev. Not sure if we plan to support Python3 for older Matomo version as well...

Findus23:
@sgiehl I think it should be possible to send data from the latest 4.x-dev log analytics script to a 3.x Matomo instance, so one can use python3 with Matomo 3.

Issue #215, opened 2018-06-01
Add an argument to select the PHP binary/version to use


domsson:
We're trying to run the `import_logs.py` on a server with several PHP versions installed, but where the `php` command (`/usr/bin/php`) is a symlink to `php4.4` and we do not have permissions to change that. We can easily run individual PHP scripts with other versions by running them with `php7.1 name-of-script.php` instead of using just `php`, but when using `import_logs.py`, it seems to run `php`, which gives the old version and hence leads to errors, aborting the script.

I looked through the command line arguments, hoping we could specify the PHP version to use, but failed to find one. Would it be possible to add such an argument? Or maybe there is a different way of achieving all this that I currently fail to see. Cheers!

fdellwing:
Please test if changing the `phpBinary` in line 1041 to `php7.1` solves the issue.

domsson:
Unaltered, the script gives this error message:

    Fatal error: `php /path-to-webroot/htdocs/matomo/misc/cron/updatetoken.php --piwik- 
       domain=sub.main.tld` failed with error: .
    Reponse code was: 255. You can alternatively run the importer using the --login and --password option

When changing `phpBinary` to `php7.1` as described above, this happens:

    Traceback (most recent call last):
      File "matomo/misc/log-analytics/import_logs.py", line 2612, in <module>
        config.init_token_auth()
      File "matomo/misc/log-analytics/import_logs.py", line 1087, in init_token_auth
        self.options.matomo_token_auth = self._get_token_auth()
      File "matomo/misc/log-analytics/import_logs.py", line 1072, in _get_token_auth
        credentials = open(filename, 'r').readline()
    IOError: [Errno 2] No such file or directory: 'X-Powered-By: PHP/7.1.17\r\nContent-type: text/html; 
    charset=UTF-8\r\n\r\n/path-to-webroot/htdocs/matomo/tmp/cache/token.php'

However, I think that's again due to the interpreter. Changing `phpBinary` to `php7.1-cli` yields this:

    0 lines parsed, 0 lines recorded, 0 records/sec (avg), 0 records/sec (current)
    Parsing log logs/access.log.16.gz...
    Fatal error: the selected log format doesn't include the hostname: you must specify the Matomo site 
    ID with the --idsite argument

Which, I assume, means that the script works fine, and I'm now facing an unrelated error.

[...]

PR #244, opened 2019-04-08
Added new command line option '--php-binary'

domsson:
To address [issue 215](https://github.com/matomo-org/matomo-log-analytics/issues/215), I've added a new command-line option, `--php-binary`, that let's the user specify the PHP binary to use. It defaults to `php`, which is what the `phpBinary` variable was hardcoded to before. In other words, if the option is not given, the script should behave exactly as before.


import_logs.py
```diff
@@ -1042,7 +1046,7 @@ def _get_token_auth(self):
                     '../../misc/cron/updatetoken.php'),
             )
 
-            phpBinary = 'php'
+            phpBinary = config.options.php_binary
```
sgiehl, line 1082: might be useful to not do the "windows-only" check some lines below if the `phpBinary` option differs from default. Otherwise it might happen that the script fatals if the `where php.exe` does not return anything but `phpBinary` option was actually set correct...

sgiehl:
I've updated this PR to make it compatible with the latest changes on 4.x-dev branch.
@diosmosis could you do a quick review as well and merge if everything looks fine?

ryrun:
I've updated to 4.2.0 today. It seems the parameter is still missing in the official build, but its mentioned in the changelog: https://matomo.org/changelog/matomo-4-2-0/ What happend?

sgiehl:
Guess we forgot to update the submodule before building the release 🤔 🙈

ryrun:
Ok, no problem, i'll wait for the next release :)

sgiehl:
Sorry for the inconvenience nevertheless. I'll later build a github action to do that automatically, so we don't forget that in the future again.

PR #246, opened 2019-06-06
import logs with client certificates (with matomo with force_ssl = 1)

leftmostlane:
Added the ability to the import logs script to import logs when force_ssl is turned on in Matomo.  
Calling import_logs.py with a client certificate activates the ssl compatibility.


sgiehl:
Actually your PR only changes a line-ending. Did you maybe forget to add your changes? If so feel free to create a new PR with the correct changes. Cheers

PR #251, opened 2019-11-23
Clarified usage of --token-auth=<SECRET>

cedwardsmedia:
When using this script with a Matomo install on a different
machine, one must include the --token-auth=<SECRET> parameter to
successfully send the logs to Matomo. This was not previously
mentioned in this particular section and is easy to overlook when
not reading about the other methods described herein.

I personally spent way too much time trying to solve this issue until
another person on irc pointed out this flag. I figured this minor edit
will save someone else from the confusion and extra research.


tsteur:
cheers 👍

PR #256, opened 2020-02-05
import logs with client certificates (matomo with require ssl turned on)

ghost:
import logs with client certificates (with matomo with force_ssl = 1)


ghost:
This is a follow up to: https://github.com/matomo-org/matomo-log-analytics/pull/246

PR #266, opened 2020-06-08
add Zabbix and GoogleStackdriverMonitoring agents

Neo2SHYAlien:
- Zabbix is one of the best open source monitoring systems. Zabbix-agent
use "Zabbix x.x.x" (version) for http/https checks
For example: "HEAD / HTTP/1.1" 200 - "-" "Zabbix 4.4.6"
- GoogleStackdriverMonitoring - Google Cloud's operations suite is
designed to monitor, troubleshoot, and improve cloud infrastructure,
software, and application performance.
For example: "GET / HTTP/2.0" 200 5683 "-"
"GoogleStackdriverMonitoring-UptimeChecks(https://cloud.google.com/monitoring)"

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -111,6 +111,8 @@ EXCLUDED_USER_AGENTS = (
     'voilabot',
     'yahoo',
     'yandex',
+    'zabbix',
+    'googlestackdrivermonitoring',
 )
 
 MATOMO_DEFAULT_MAX_ATTEMPTS = 3


sgiehl:
@Neo2SHYAlien thanks for creating the PR. We actually do not want to maintain another "big list" of bots in log analytics. It's fine to have some basic detections, so a majority of bot requests won't be sent to Matomo at all, but all others should be detected by Matomo.  So if have some other special bots, that are still tracked in Matomo feel free to create an issue here: https://github.com/matomo-org/device-detector

PR #269, opened 2020-06-22
Merge 3.x into 4 (for review only)

diosmosis:


import_logs.py
```diff
@@ -436,20 +473,27 @@ def get(self, key):
     r'\s+"(?P<referrer>.*?)"\s+"(?P<user_agent>.*?)"'
 )
 
+# We need to double escape the word-boundry to work correctly, because https://docs.python.org/2/reference/lexical_analysis.html#string-literals
```
sgiehl, line 476: Guess it's the same for Python 3 ? Maybe we can update the link to point to Python 3 docs instead then...

sgiehl (approved):
Looks good otherwise

import_logs.py
diosmosis: Actually I removed this bit for python 3, will remove the comment as well.

diosmosis:
merged manually

PR #275, opened 2020-07-24
Fix importing logs using stdin

### LINKED ISSUE #273: When input is stdin ('-'), import_logs.py doesn't work
With the latest changes (I think commit: https://github.com/matomo-org/matomo-log-analytics/commit/58e162ea9c10d3d2e4859e869ea586edf818eaa6) I am unable to import logs from syslog using the recommended way (https://github.com/matomo-org/matomo-log-analytics/tree/3.x-dev#matomosh-syslog-ng-version)

The input file for `ìmport_logs.py` is `-`, but I think it is removed in the following line: https://github.com/matomo-org/matomo-log-analytics/blob/3.x-dev/import_logs.py#L952 (`all_filenames = all_filenames + glob.glob(self.filename)`)

sgiehl:
fixes #273

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -949,7 +949,10 @@ class Configuration(object):
 
         all_filenames = []
         for self.filename in self.filenames:
-            all_filenames = all_filenames + glob.glob(self.filename)
+            if self.filename == '-':
+                all_filenames.append(self.filename)
+            else:
+                all_filenames = all_filenames + glob.glob(self.filename)
         self.filenames = all_filenames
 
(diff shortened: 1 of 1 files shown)


mattab:
Can we add a test for this case so it won't regress again in the future?

sgiehl:
@mattab I'm not that deep into python. Would need to investigate if it's possible to "overwrite" stdin in tests

diosmosis:
If it's still important to add tests: https://stackoverflow.com/questions/6271947/how-can-i-simulate-input-to-stdin-for-pyunit

PR #277, opened 2020-07-29
Add "--use-http-x-forwarded-for" option for NGINX

adityapahuja:
This pull request

- updates the script to pick up an ip address from the `http_x_forwarded_for` field instead of the `ip` field in NGINX JSON formatted log file provided that the "--use-http-x-forwarded-for" option is used.  If the `http_x_forwarded_for` field contains "-" or is missing, the script will pick up the ip address from the `ip` field. This option is useful if the `ip` field does not contain client's ip address (e.g. using NGINX as a reverse proxy).
- updates README file to include a field `http_x_forwarded_for` in NGINX log file and information about the usage of "--use-http-x-forwarded-for" option.
- adds two test cases to test the "--use-http-x-forwarded-for" option to ensure that the script works properly.

This feature is useful for replaying requests coming from NGINX reverse proxy to Matomo.


kutovoys:
@adityapahuja Thank you! That helped me.

import_logs.py
```diff
@@ -891,6 +900,11 @@ def _create_parser(self):
             default=False,
             help="Do not verify the SSL / TLS certificate when contacting the Matomo server. This is the default when running on Python 2.7.8 or older."
         )
+        option_parser.add_option(
```
kutovoys, line 903: For new versions it will be 
`parser.add_argument(`

kutovoys (changes requested):

PR #280, opened 2020-10-03
Support both AWS Application and Elastic Load Balancer log formats

### LINKED ISSUE #209: Support for importing AWS ELB access logs analytics (Application Load Balancer)
Note, AWS Elastic Load Balancing has three types of load balancers: Application Load Balancers, Network Load Balancers, and Classic Load Balancers.  It appears that, currently, Matomo only supports the Classic Load Balancer format.  (see this closed ticket: https://github.com/matomo-org/matomo-log-analytics/issues/131)

The Application Load Balancer format is different.
(https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-log-entry-format)

Here is an example from that page that is simiarl to the logs we are trying to import into Matomo 3.3:

```
h2 2016-08-10T00:10:33.145057Z app/my-loadbalancer/50dc6c495c0c9188 
10.0.1.252:48160 10.0.0.66:9000 0.000 0.002 0.000 200 200 5 257 
"GET https://10.0.2.105:773/ HTTP/2.0" "curl/7.46.0" ECDHE-RSA-AES128-GCM-SHA256 TLSv1.2
arn:aws:elasticloadbalancing:us-east-2:123456789012:targetgroup/my-targets/73e2d6bc24d8a067
"Root=1-58337327-72bd00b0343d75b906739c42" - -
```
Currently, when I try to import these logs with "--log-format-name=elb", nothing gets imported.  If I leave out that command line option let import_logs.py try to figure it out on its own, I get the following error:

> Fatal error: cannot automatically determine the log format using the first 100000 lines of the log file. 
> Maybe try specifying the format with the --log-format-name command line argument.

Would be wonderul for Matomo to get full AWS ELB support.

tabascoterrier:
To address issue #209 , I've amended the ELB format regex to support both the previous Elastic Load Balancer and the newer Application Load Balancer formats.

Just adding a simple optional non-capturing group to the start of the regex seems to be enough.

Fixes #209


diosmosis:
Thanks @tabascoterrier for the contribution!

PR #281, opened 2020-10-15
Fix for "AttributeError: 'list' object has no attribute 'get'" in

hvanderheide:
StaticResolver

On success call_api() returns a list of dicts which StaticResolver
doesn't seem to expect. On failure just the single dict is returned.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -1649,6 +1649,11 @@ class StaticResolver:
         site = matomo.call_api(
             'SitesManager.getSiteFromId', idSite=self.site_id
         )
+
+        # On success call_api returns a list of dicts
+        if isinstance(site, list):
+            site = site[0]
+
         if site.get('result') == 'error':
             fatal_error(
                 "cannot get the main URL of this site: %s" % site.get('message')


diosmosis:
This is currently tested via https://github.com/matomo-org/matomo/blob/4.x-dev/tests/PHPUnit/System/ImportLogsTest.php, I would expect the test to fail if this error was always the case. What version of Matomo / PHP / python are you using?

The result of SitesManager.getSiteFromId looks like this for me:

```
{'idsite': '1', 'name': 'Piwik test', 'main_url': 'http://piwik.net', 'ts_created': '2012-08-08 11:22:33', 'ecommerce': '0', 'sitesearch': '1', 'sitesearch_keyword_parameters': '', 'sitesearch_category_parameters': '', 'timezone': 'UTC', 'currency': 'USD', 'exclude_unknown_urls': '0', 'excluded_ips': '', 'excluded_parameters': '', 'excluded_user_agents': '', 'group': '', 'type': 'website', 'keep_url_fragment': '0', 'creator_login': 'superUserLogin', 'timezone_name': 'UTC', 'currency_name': 'US Dollar'}
```

(a dict, not a list)

Prinzhorn:
I had to apply this fix for the script to run at all. I'm running an empty plain Matomo via the official [docker-compose](https://github.com/matomo-org/docker/tree/d4744fbab41860bef1aacf5a274290a214ea84dc/.examples/apache). Should make it easy to reproduce.

```sh
python3 import_logs.py \
  --url='http://localhost:8080' \
  --token-auth=1234 \
  --idsite=1 \
  --recorders=4 \
  log-file
```

sgiehl:
@hvanderheide @Prinzhorn  I assume you are both using the 4.x-dev branch of log-analytics with a Matomo 3.x release.
The response of `SitesManager.getSiteFromId` changed for Matomo 4. So when using Matomo 4 (which will be released soon), the log import should work without that fix.

Prinzhorn:
@sgiehl yes, it was just yesterday that I realized the default branch was configured to "4.x-dev". I just cloned the repo and started using it. Didn't realize the requirements said "Matomo >= 4.0.0". Actually I don't even know what Matomo version I'm using and that 4 isn't even released yet. I just fired up the docker-compose and started experimenting. Turns out I'm on the best version you'll probably ever release (3.14.1) unless you make it to 3.14.15

Could the very first step of the import script be to fetch the Matomo version and check if you're using the right script version?

hvanderheide:
@sgiehl thanks, until now I hadn't noticed. I'll close the PR

PR #283, opened 2020-11-04
Extra host from CloudFront x-host-header

cristiklein:
CloudFront logs do include the host. However, without this patch,
`import_logs.py` will fail with the following error:

```
Fatal error: the selected log format doesn't include the hostname: you must specify the Matomo site ID with the --idsite argument
```


sgiehl (approved):

sgiehl:
According the aws documentation `x-host-header` contains:

```
The value that the viewer included in the Host header of the request. If you’re using the CloudFront domain name in your object URLs (such as d111111abcdef8.cloudfront.net), this field contains that domain name. If you’re using alternate domain names (CNAMEs) in your object URLs (such as www.example.com), this field contains the alternate domain name.

If you’re using alternate domain names, see cs(Host) in field 7 for the domain name that is associated with your distribution.
```

So should imho be correct to use that as host

PR #288, opened 2020-11-17
Fix importing compressed log files

### LINKED ISSUE #287: Import *.gz files: UnboundLocalError: local variable 'file' referenced before assignment
From what I understand the script should automatically detect and unzip these? It doesn't like my `*.gz` Apache logs. Can be reproduced with an empty gzipped file (literally `gzip empty` with a 0 byte `empty` file).

```sh
python3 import_logs.py \
  --dry-run \
  --url='http://localhost:8080' \
  --token-auth=1234 \
  --idsite=1 \
  empty.gz
```

```
Parsing log empty.gz...
  File "import_logs.py", line 2674, in <module>
    main()
  File "import_logs.py", line 2640, in main
    parser.parse(filename)
  File "import_logs.py", line 2378, in parse
    data = file.read(100)
UnboundLocalError: local variable 'file' referenced before assignment
```

Edit: I also had to apply #281 before the script even ran

sgiehl:
fixes #287


tests/test_main.py
```diff
@@ -1163,3 +1163,41 @@ def test_matomo_error_construct():
...
+    import_logs.parser = import_logs.Parser()
+    import_logs.parser.parse(file_)
+
+    hits = [hit.__dict__ for hit in Recorder.recorders]
+
+    assert hits[0]['ip'] == '1.2.3.4'
+    assert hits[0]['path'] == '/'
+    assert hits[0]['status'] == '301'
+    assert hits[0]['length'] == 368
+    assert hits[0]['userid'] == 'theuser'
```
diosmosis: think we can change some of the data in the bz2 test? just to show that the global state was reset properlly

diosmosis (commented):
lgtm, left one comment

tests/test_main.py
sgiehl: sure. makes sense. will change that

sgiehl: @diosmosis done. Is that good to merge then?

PR #292, opened 2020-11-18
explicitly sort glob results for an implicit progress bar

### LINKED ISSUE #291: Explicitly sort globs for implicit progress bar
I have 122k log files in a folder (CloudFront logs). They are named like this

```
DISTRIBUTION-ID.2017-05-10-19.some-token
```

The files seem to be processed in a seemingly random order. Python docs regarding `glob.glob` say

> Whether or not the results are sorted depends on the file system

Would it be possible we explicitly sort the glob result? Right now I have absolutely no sense of progress while the import is happening. If they would be processed in a natural order I could look at the import and be like "oh it's processing december 2018 right now, almost two years to go"

Prinzhorn:
fixes #291

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,3 @@
+# Next
+
+- Process file globs in a sorted order (#291)
--- a/import_logs.py
+++ b/import_logs.py
@@ -938,7 +938,7 @@ class Configuration:
             if self.filename == '-':
                 all_filenames.append(self.filename)
             else:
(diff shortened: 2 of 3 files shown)


Findus23:
Seem reasonable to me as unsorted globs can be annoying and sorting the list of filenames should never have a relevent performance impact.

sgiehl (approved):

Prinzhorn:
Thanks!

PR #296, opened 2020-12-21
--force-all-periods is not supported and --force-date-last-n only applies to new segments

### LINKED ISSUE #16931: import_logs.py instructs to use --force-all-periods but core:archive says that switch DNE
Output from import_logs.py:

<pre>Processing your log data
------------------------

    In order for your logs to be processed by Matomo, you may need to run the following command:
     ./console core:archive --force-all-websites --force-all-periods=315576000 --force-date-last-n=1000 --url='URL'</pre>

Corresponding console output:

<pre>ERROR [2020-12-10 03:09:40] 5554  Uncaught exception: /var/www/matomo/vendor/symfony/console/Symfony/Component/Console/Input/ArgvInput.php(213): The "--force-all-periods" option does not exist.
The "--force-all-periods" option does not exist.

                                                    
  [RuntimeException]                                
  The "--force-all-periods" option does not exist. 
</pre>

diosmosis:
`--force-all-periods` is not supported and `--force-date-last-n` only applies to new segments.

Fixes https://github.com/matomo-org/matomo/issues/16931

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -1276,7 +1276,7 @@ Processing your log data
 ------------------------
 
     In order for your logs to be processed by Matomo, you may need to run the following command:
-     ./console core:archive --force-all-websites --force-all-periods=315576000 --force-date-last-n=1000 --url='%(url)s'
+     ./console core:archive --force-all-websites --url='%(url)s'
 ''' % {
 
     'count_lines_recorded': self.count_lines_recorded.value,


diosmosis:
Replace `--force-date-last-n` everywhere it was remaining (in regards to matomo 4.x docs) on matomo.org.

sgiehl (approved):

PR #305, opened 2021-01-24
Fix infinite loop in format detection

### LINKED ISSUE #304: Infinite loop in log format detection
`import_logs.py` gets stuck in an infinite loop in log format detection in this example (Matomo v4.1.1):

```shell
echo '# test' > somefile
python3 misc/log-analytics/import_logs.py --url=http://localhost/ somefile
```

It tries to read lines at end of file in this loop:
https://github.com/matomo-org/matomo-log-analytics/blob/78613990160194afbd2d18792858c2d154480605/import_logs.py#L291-L295

sth:
Fixes #304 by breaking out of the loop when no more data can be read.

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -424,7 +424,11 @@ class W3cExtendedFormat(RegexFormat):
         # if we're reading from stdin, we can't seek, so don't read any more than the Fields line
         header_lines = []
         while fields_line is None:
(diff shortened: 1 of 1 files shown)


sth:
I don't really see how this test failure would be related to the change.

sgiehl (approved):

PR #310, opened 2021-04-14
Certificate Arguments for Secure Web Servers

ghost:
Adds arguments to specify public, private and trust certificates for importing logs onto servers where SSL is enabled and client certificates are required.


ghost:
@diosmosis or @sgiehl  please take a look at this, thanks.

PR #312, opened 2021-05-26
Reports signature files as downloads

Pierre-Lannoy:
Allows to reports downloaded signature files for what they are:  downloads!


sgiehl:
Hi @Pierre-Lannoy. Thanks for creating this pull request. Did you open the request against the master branch on purpose?
The master branch actually holds an outdated version of the script that is not compatible with Matomo 4 anymore.
If you want to fix this in the latest release you need to target the (default) branch `4.x-dev`.

Pierre-Lannoy:
Aha. Thanks. There's no *true* reason to have done it on master branch... Sorry for that; I will create it on `4.x-dev` branch.

PR #313, opened 2021-05-26
Reports signature files as downloads

Pierre-Lannoy:
Allows to reports downloaded signature files for what they are: downloads!


Pierre-Lannoy:
Thank you so much for the merge... :)

Issue #286, opened 2020-11-07
TypeError: the JSON object must be str, not 'bytes' in import_logs.py


1000i100:
When i run `./misc/log-analytics/import_logs.py` following https://github.com/matomo-org/matomo-log-analytics#how-do-i-setup-nginx-to-directly-import-to-matomo-via-syslog
( and after update like describe here :  #285 )

Here is what i get : 
```
Traceback (most recent call last):
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "******/misc/log-analytics/import_logs.py", line 1849, in _run_bulk
    self._record_hits(hits)
  File ""******/misc/log-analytics/import_logs.py", line 1995, in _record_hits
    'requests': [self._get_hit_args(hit) for hit in hits]
  File ""******/misc/log-analytics/import_logs.py", line 1995, in <listcomp>
    'requests': [self._get_hit_args(hit) for hit in hits]
  File ""******/misc/log-analytics/import_logs.py", line 1891, in _get_hit_args
    site_id, main_url = resolver.resolve(hit)
  File ""******/misc/log-analytics/import_logs.py", line 1775, in resolve
    return self._resolve_by_host(hit)
  File ""******/misc/log-analytics/import_logs.py", line 1758, in _resolve_by_host
    site_id = self._resolve(hit)
  File ""******/misc/log-analytics/import_logs.py", line 1726, in _resolve
    res = self._get_site_id_from_hit_host(hit)
  File ""******/misc/log-analytics/import_logs.py", line 1681, in _get_site_id_from_hit_host
    url=hit.host,
  File ""******/misc/log-analytics/import_logs.py", line 1633, in call_api
    return self._call_wrapper(self._call_api, None, None, method, **kwargs)
  File ""******/misc/log-analytics/import_logs.py", line 1585, in _call_wrapper
    response = func(*args, **kwargs)
  File ""******/misc/log-analytics/import_logs.py", line 1574, in _call_api
    return json.loads(res)
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'

1365 lines parsed, 0 lines recorded, 0 records/sec (avg), 0 records/sec (current)
[... Ctrl+C to stop it ...]
[...]

PR #316, opened 2021-06-24
Add Gandi VirtualHost log format

tups:
### Description:

I suggest adding the regex for Gandi's virtualhost servers, because the log format is different.
Thank you anyway for your work

### Review

* [ ] Functional review done
* [ ] Potential edge cases thought about (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] Usability review done (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] Security review done [see checklist](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] Code review done
* [ ] Tests were added if useful/possible
* [ ] Reviewed for breaking changes
* [ ] Developer changelog updated if needed
* [ ] Documentation added if needed
* [ ] Existing documentation updated if needed

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -462,9 +462,6 @@ _NCSA_EXTENDED_LOG_FORMAT = (_COMMON_LOG_FORMAT +
 )
 
 
-_GANDI_SIMPLE_HOSTING_FORMAT = (
-    r'(?P<host>[0-9a-zA-Z-_.]*)\s(?P<ip>[a-zA-Z0-9.]*)\s(\S*)\s(?P<userid>\S*)\s\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s\((?P<generation_time_secs>[0-9a-zA-Z\s]*)\)\s"(?P<method>[A-Z]*)\s(?P<path>[^\s]*)\s([^\s]*)"\s(?P<status>[0-9]*)\s(?P<length>[^\s]*)\s"(?P<referrer>[^\s]*)"\s"(?P<user_agent>[^"]*)"'
-)
 _S3_LOG_FORMAT = (
     r'\S+\s+(?P<host>\S+)\s+\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s+(?P<ip>[\w*.:-]+)\s+'
     r'(?P<userid>\S+)\s+\S+\s+\S+\s+\S+\s+"(?P<method>\S+)\s+(?P<path>.*?)\s+\S+"\s+(?P<status>\d+)\s+\S+\s+(?P<length>\S+)\s+'
@@ -490,7 +487,7 @@ _HAPROXY_FORMAT = (
 )
 
 _GANDI_SIMPLE_HOSTING_FORMAT = (
-    r'(?P<host>[0-9a-zA-Z-_.]*)\s(?P<ip>[a-zA-Z0-9.]*)\s(\S*)\s(?P<userid>\S*)\s\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s\((?P<generation_time_secs>[0-9a-zA-Z\s]*)\)\s"(?P<method>[A-Z]*)\s(?P<path>[^\s]*)\s([^\s]*)"\s(?P<status>[0-9]*)\s(?P<length>[^\s]*)\s"(?P<referrer>[^\s]*)"\s"(?P<user_agent>[^"]*)"'
+    r'(?P<host>[0-9a-zA-Z-_.]+)\s+(?P<ip>[a-zA-Z0-9.]+)\s+\S+\s+(?P<userid>\S+)\s+\[(?P<date>.+?)\s+(?P<timezone>.+?)\]\s+\((?P<generation_time_secs>[0-9a-zA-Z\s]*)\)\s+"(?P<method>[A-Z]+)\s+(?P<path>[^\s]+)\s+([^\s]+)"\s+(?P<status>[0-9]+)\s+(?P<length>[^\s]+)\s+"(?P<referrer>[^\s]+)"\s+"(?P<user_agent>[^"]+)"'
 )
 
 FORMATS = {
--- a/import_logs.py
+++ b/import_logs.py
@@ -459,6 +459,11 @@ _COMMON_LOG_FORMAT = (
 _NCSA_EXTENDED_LOG_FORMAT = (_COMMON_LOG_FORMAT +
     r'\s+"(?P<referrer>.*?)"\s+"(?P<user_agent>.*?)"'
 )
+
+
+_GANDI_SIMPLE_HOSTING_FORMAT = (
+    r'(?P<host>[0-9a-zA-Z-_.]*)\s(?P<ip>[a-zA-Z0-9.]*)\s(\S*)\s(?P<userid>\S*)\s\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s\((?P<generation_time_secs>[0-9a-zA-Z\s]*)\)\s"(?P<method>[A-Z]*)\s(?P<path>[^\s]*)\s([^\s]*)"\s(?P<status>[0-9]*)\s(?P<length>[^\s]*)\s"(?P<referrer>[^\s]*)"\s"(?P<user_agent>[^"]*)"'
+)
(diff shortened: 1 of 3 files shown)


import_logs.py
```diff
@@ -498,7 +502,8 @@ def get(self, key):
     'elb': RegexFormat('elb', _ELB_LOG_FORMAT, '%Y-%m-%dT%H:%M:%S'),
     'nginx_json': JsonFormat('nginx_json'),
     'ovh': RegexFormat('ovh', _OVH_FORMAT),
-    'haproxy': RegexFormat('haproxy', _HAPROXY_FORMAT, '%d/%b/%Y:%H:%M:%S.%f')
+    'haproxy': RegexFormat('haproxy', _HAPROXY_FORMAT, '%d/%b/%Y:%H:%M:%S.%f'),
+    'gandi': RegexFormat('gandi', _GANDI_FORMAT, '%d/%b/%Y:%H:%M:%S')
```
sgiehl: This needs to match the name as above: `_GANDI_SIMPLE_HOSTING_FORMAT `

sgiehl (changes requested):
Hi @tups 
Thanks for the contribution. The current implementation seems not to work due to a typo in the format name.
Also it would be awesome if you could add a test for the new format with a small example log. See other tests in https://github.com/matomo-org/matomo-log-analytics/blob/4.x-dev/tests/test_main.py

import_logs.py
tups: I just corrected this error and added the tests for the "Gandi" format

import_logs.py
```diff
@@ -1527,9 +1538,10 @@ def _call(self, path, args, headers=None, url=None, data=None):
             self.RedirectHandlerWithLogging(),
             urllib.request.HTTPSHandler(**https_handler_args))
         response = opener.open(request, timeout = timeout)
+        encoding = response.info().get_content_charset('utf-8')
         result = response.read()
         response.close()
-        return result
+        return result.decode(encoding)
```
sgiehl, line 1541: why exactly are those change required?

tups: I had the same problem as this issue : https://github.com/matomo-org/matomo-log-analytics/issues/286

Here is the result when I was running on the server:

```hosting-user@server:/srv/data$ python3.5 ./web/vhosts/stats.example.com/htdocs/misc/log-analytics/import_logs.py --url=https://stats.example.com --debug --debug-tracker --token-auth=f1f1f1f1f11f1f1f1f1f1f1f1f1f1 --idsite=1 --hostname=www.example.com --enable-http-errors --enable-http-redirects --enable-static --enable-bots --log-format-name=gandi ./home/access.log
2021-06-25 08:56:50,923: [DEBUG] Accepted hostnames: www.example.comstatic --ena                                              2021-06-25 08:56:50,924: [DEBUG] Matomo Tracker API URL is: https://stats.example.com
2021-06-25 08:56:50,924: [DEBUG] Matomo Analytics API URL is: https://stats.example.com
2021-06-25 08:56:50,924: [DEBUG] Authentication token token_auth is: f1f1f1f1f11f1f1f1f1f1f1f1f1f1
2021-06-25 08:56:50,924: [DEBUG] Resolver: static
  File "./web/vhosts/stats.example.com/htdocs/misc/log-analytics/import_logs.py", line 2687, in <module>
    resolver = config.get_resolver()
  File "./web/vhosts/stats.example.com/htdocs/misc/log-analytics/import_logs.py", line 1129, in get_resolver
    return StaticResolver(self.options.site_id)
  File "./web/vhosts/stats.example.com/htdocs/misc/log-analytics/import_logs.py", line 1663, in __init__
    'SitesManager.getSiteFromId', idSite=self.site_id
  File "./web/vhosts/stats.example.com/htdocs/misc/log-analytics/import_logs.py", line 1646, in call_api
    return self._call_wrapper(self._call_api, None, None, method, **kwargs)
  File "./web/vhosts/stats.example.com/htdocs/misc/log-analytics/import_logs.py", line 1598, in _call_wrapper
  File "./web/vhosts/stats.example.com/htdocs/misc/log-analytics/import_logs.py", line 1586, in _call_api
    return json.loads(res)
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```
I tried the different proposals for the issue, without success.
I ended up finding this solution, which converts the JSON string well to the format wanted by json.loads at line 1586

It would indeed be interesting to try it on another environment than the one I used

import_logs.py
```diff
@@ -459,6 +459,11 @@ def get(self, key):
 _NCSA_EXTENDED_LOG_FORMAT = (_COMMON_LOG_FORMAT +
     r'\s+"(?P<referrer>.*?)"\s+"(?P<user_agent>.*?)"'
 )
+
+
+_GANDI_SIMPLE_HOSTING_FORMAT = (
+    r'(?P<host>[0-9a-zA-Z-_.]*)\s(?P<ip>[a-zA-Z0-9.]*)\s(\S*)\s(?P<userid>\S*)\s\[(?P<date>.*?)\s+(?P<timezone>.*?)\]\s\((?P<generation_time_secs>[0-9a-zA-Z\s]*)\)\s"(?P<method>[A-Z]*)\s(?P<path>[^\s]*)\s([^\s]*)"\s(?P<status>[0-9]*)\s(?P<length>[^\s]*)\s"(?P<referrer>[^\s]*)"\s"(?P<user_agent>[^"]*)"'
+)
```
diosmosis: Hi @tups, is there a reason some groups here aren't named groups? And is it possible for the whitespace parts to use `\s+` like the other regexes, as well as the fields we don't care about be parsed with `\S+`?

tups: Hi, I just checked, it is actually not necessary to capture this. It corresponds to the format %l of the apache log: https://httpd.apache.org/docs/2.4/fr/logs.html
I just improved the RegEx with your recommendations

import_logs.py
```diff
@@ -483,6 +486,10 @@ def get(self, key):
     r'.*:\ (?P<ip>[\w*.]+).*\[(?P<date>.*)\].*\ (?P<status>\b\d{3}\b)\ (?P<length>\d+)\ -.*\"(?P<method>\S+)\ (?P<path>\S+).*'
 )
 
+_GANDI_SIMPLE_HOSTING_FORMAT = (
+    r'(?P<host>[0-9a-zA-Z-_.]+)\s+(?P<ip>[a-zA-Z0-9.]+)\s+\S+\s+(?P<userid>\S+)\s+\[(?P<date>.+?)\s+(?P<timezone>.+?)\]\s+\((?P<generation_time_secs>[0-9a-zA-Z\s]*)\)\s+"(?P<method>[A-Z]+)\s+(?P<path>[^\s]+)\s+([^\s]+)"\s+(?P<status>[0-9]+)\s+(?P<length>[^\s]+)\s+"(?P<referrer>[^\s]+)"\s+"(?P<user_agent>[^"]+)"'
+)
```
diosmosis, line 491: One more change than it should be good to merge: the `[^\s]+` parts can be replaced with just `\S+`, correct?

tups: It is much the same thing :
`\s`: any whitespace characters
`\S`: any non-whitespace characters

So, `[^\s]` (all characters except any whitespace characters) is equal to `\S` (any non-whitespace characters).

But if you really want it, I change it :)

diosmosis: Hi @tups can you please change it to keep it consistent with the other regexes? Thanks!

tups: It is done

diosmosis:
Thanks for the PR @tups and thanks for applying all the requested review feedback!

import_logs.py
```diff
@@ -1527,9 +1538,10 @@ def _call(self, path, args, headers=None, url=None, data=None):
             self.RedirectHandlerWithLogging(),
             urllib.request.HTTPSHandler(**https_handler_args))
         response = opener.open(request, timeout = timeout)
+        encoding = response.info().get_content_charset('utf-8')
```
metalocator, line 1538: This line and the result.decode below caused the issue described here:

https://forum.matomo.org/t/log-file-importer-error-utf-8-codec-cant-decode-byte-0x80-in-position-10-invalid-start-byte/43169

PR #329, opened 2022-01-21
test in newer python versions

Findus23:
Now that Python 3.9 and 3.10 are stable releases, I think it makes sense to also run tests for them just to be safe.

The new two test results:
https://app.travis-ci.com/github/matomo-org/matomo-log-analytics/jobs/556495978
https://app.travis-ci.com/github/matomo-org/matomo-log-analytics/jobs/556495979


sgiehl:
makes sense.

PR #358, opened 2023-08-18
Add Traefik json log support

sethlinnkuleuven:
### Description:
This builds on pull request #322 (which can be deleted if this is accepted) This pull request adds support for the Traefik json log format, not to be confused with the Traefik common log.

What was done:
1. Expanded import_logs.py to auto recognize and import Traefik json log files.
2. Modified the NGINX json support, made the 'check_format_line' function more specific. It was to generic. All json formatted logs were autodetected as NGINX logs.
3. Added tests for Traefik log importing.


sgiehl:
@sethlinnkuleuven I've fixed the test action to run correctly again. Would you mind pulling in the latest changes, so we can see if they pass here?

sethlinnkuleuven:
Ok, I pulled your changes, lets have another run. Thanks!

sgiehl (approved):
Looks good to me. Tests are passing, so should be good to go.
Thanks for the contribution @sethlinnkuleuven

PR #374, opened 2024-06-20
caddy_json: fix user_agent NoneType

simon04:
```
  File "/var/www/analytics/misc/log-analytics/import_logs.py", line 2825, in <module>
    main()
  File "/var/www/analytics/misc/log-analytics/import_logs.py", line 2791, in main
    parser.parse(filename)
  File "/var/www/analytics/misc/log-analytics/import_logs.py", line 2636, in parse
    if hit.user_agent.startswith('"'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```


sgiehl:
should be fixed with https://github.com/matomo-org/matomo-log-analytics/pull/381 already

PR #379, opened 2025-03-30
Add support for varnishncsa format.

lapo-luchini:
It is similar to ncsa_extended but has protocol and hostname inside the request.
https://varnish-cache.org/docs/7.5/reference/varnishncsa.html#notes


sgiehl:
@lapo-luchini Thanks for creating this PR. Would you be able to adjust and add some tests for it?

PR #394, opened 2026-05-22
[Patch available] --enable-bots: Support Custom Dimensions as replace…

### LINKED ISSUE #392: AI bot detection not working properly with log file imports
Hi there,

I'm conducting research for my Bachelor's thesis on AI bot traffic at the University of Tübingen (Germany). Due to privacy requirements (GDPR), we're using Matomo in offline mode with daily log imports via import_logs.py.

**Issue:**

I'm trying to analyze AI bot and assistant traffic patterns, but the detection seems incomplete. The AI Assistants dashboard shows only 453 AI agent visits out of ~500,000 total visits per day, which appears significantly underreported.

**What I've tried:**

Manual rotation through log files with HTTP Tracking API requests
However, with 2-4 million entries per log file per day, the API consistently times out or fails

**Questions:**

Is AI bot detection known to have limitations with the log import method (import_logs.py)?
Are there specific parameters or configurations that improve AI bot detection for log imports?
Would a custom post-processing solution (e.g., analyzing logs before import) be more reliable?

**Environment:**

Matomo 5.8.0
Import method: misc/log-analytics/import_logs.py
Log volume: 2-4 million requests/day
Using --enable-bots flag

Any guidance would be greatly appreciated!
Best regards,
Emanuel Ganß

ManuShift:
Fixes #392

…ment for deprecated Custom Variables

Add--enable-bots does not work on Matomo 5.10 (Custom Variables deprecated)

On Matomo 5.10 the CustomVariables plugin is no longer part of core and must be installed separately from the Marketplace. When the plugin is not available, --enable-bots silently fails to store any bot information.. the flag still includes bot traffic, but the _cvar data is discarded by the tracker.

Proposed fix

Extend --enable-bots to optionally accept a Custom Dimension ID:

--enable-bots → existing behavior, backward compatible --enable-bots=1 → stores the classified bot name in Custom Dimension 1 (recommended for Matomo 5.*) This adds a built-in bot classification table (GPTBot, Bingbot, ClaudeBot, SemrushBot, etc.) and uses dimension{N} in the tracker request instead of _cvar.

Non-bot visits receive Not a Bot in the dimension, allowing easy segmentation.

Motivation

Custom Dimensions are the official replacement for Custom Variables in Matomo 5 (see deprecation notice). The log importer should reflect this.ed new bot user agents and updated bot tracking logic.

### Description

Please include a description of this change and which issue it fixes. If no issue exists yet please include context and what problem it solves.

### Checklist

- [✔/✖/NA] I have understood, reviewed, and tested all AI outputs before use
- [✔/✖/NA] All AI instructions respect security, IP, and privacy rules

### Review

- [x] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
- [x] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behaviour of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
- [x] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
- [x] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
- [x] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
- [x] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
- [x] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
- [x] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
- [x] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
- [x] [New documentation added or existing documentation updated](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)

### DIFF
--- a/import_logs.py
+++ b/import_logs.py
@@ -957,13 +957,16 @@ class Configuration:
         )
         parser.add_argument(
             '--enable-bots', dest='enable_bots',
-            nargs='?', const=True, default=None,
+            action='store_true', default=False,
+            help="Track bots. All bot visits will have a Custom Variable set with name='Bot' and value='$Bot_user_agent_here$'"
+        )
+        parser.add_argument(
+            '--bot-custom-dimension', dest='bot_custom_dimension',
+            type=int, default=None,
             metavar='DIMENSION_ID',
-            help="Track bots. When used without a value (--enable-bots), bot visits will have a Custom Variable "
-                 "set with name='Bot' and value='$Bot_user_agent_here$' (requires the CustomVariables plugin). "
-                 "When used with a Custom Dimension ID (--enable-bots=1), the bot name is stored in that dimension "
-                 "instead, which is the recommended approach for Matomo 5+ where Custom Variables are deprecated. "
-                 "Non-bot visits receive 'Not a Bot' in the dimension."
+            help="Store the bot name in the given Custom Dimension ID instead of a Custom Variable "
+                 "(recommended for Matomo 5+ where Custom Variables are deprecated). "
+                 "Requires --enable-bots. Non-bot visits receive 'Not a Bot' in the dimension."
         )
         parser.add_argument(
             '--enable-http-errors', dest='enable_http_errors',
@@ -2338,16 +2341,8 @@ class Recorder:
 
         # handle bot tracking
         if config.options.enable_bots:
-            if config.options.enable_bots is True:
-                dimension_id = None
-            else:
-                try:
-                    dimension_id = int(config.options.enable_bots)
-                except (TypeError, ValueError):
-                    dimension_id = None
-
-            if dimension_id is not None:
-                dim_key = 'dimension%d' % dimension_id
+            if config.options.bot_custom_dimension is not None:
+                dim_key = 'dimension%d' % config.options.bot_custom_dimension
                 if hit.is_robot:
                     hit.args[dim_key] = _get_bot_name_for_custom_dimension(hit.user_agent)
                 else:
@@ -2386,7 +2381,7 @@ class Recorder:
         if hit.is_download:
             args['download'] = args['url']
 
-        if config.options.enable_bots:
+        if config.options.enable_bots or config.options.bot_custom_dimension is not None:
             args['bots'] = '1'
 
         if hit.is_error or hit.is_redirect:
@@ -2617,7 +2612,7 @@ class Parser:
         user_agent = hit.user_agent.lower()
         for s in itertools.chain(EXCLUDED_USER_AGENTS, config.options.excluded_useragents):
             if s in user_agent:
-                if config.options.enable_bots:
+                if config.options.enable_bots or config.options.bot_custom_dimension is not None:
                     hit.is_robot = True
                     return True
                 else:
--- a/import_logs.py
+++ b/import_logs.py
@@ -84,35 +84,148 @@ DOWNLOAD_EXTENSIONS = set((
 # user agents must be lowercase
 EXCLUDED_USER_AGENTS = (
     'adsbot-google',
+    'amazonbot',
+    'anthropic-ai',
+    'applebot',
+    'archive.org_bot',
+    'aranet-searchbot',
     'ask jeeves',
+    'awariobot',
     'baidubot',
+    'bitsightbot',
     'bot-',
     'bot/',
+    'bytespider',
+    'ccbot',
     'ccooter/',
+    'chatgpt-user',
+    'claude-web',
+    'claudebot',
+    'codabot',
+    'cohere-ai',
     'crawl',
     'curl',
+    'dataforseobot',
+    'diffbot',
+    'duckassistbot',
     'echoping',
     'exabot',
+    'facebookbot',
     'feed',
+    'google-extended',
     'googlebot',
+    'googleother',
     'ia_archiver',
     'java/',
     'libwww',
     'mediapartners-google',
+    'meta-externalagent',
     'msnbot',
     'netcraftsurvey',
+    'oai-searchbot',
     'panopta',
+    'petalbot',
+    'perplexitybot',
     'pingdom.com_bot_',
+    'qwantbot',
     'robot',
+    'seznambot',
+    'sleepbot',
     'spider',
     'surveybot',
+    'thinkers-bot',
+    'tiktokspider',
+    'trendictionbot',
     'twiceler',
     'voilabot',
     'yahoo',
     'yandex',
+    'youbot',
     'zabbix',
     'googlestackdrivermonitoring',
 )
+
+# Known bot user agent patterns mapped to display names, used with --bot-custom-dimension.
+# Patterns must be lowercase substrings of the user agent string.
+BOT_CUSTOM_DIMENSION_NAMES = [
+    # AI / LLM bots
+    ('GPTBot',               'gptbot'),
+    ('ChatGPT-User',         'chatgpt-user'),
+    ('OAI-SearchBot',        'oai-searchbot'),
+    ('ClaudeBot',            'claudebot'),
+    ('Claude-Web',           'claude-web'),
+    ('Anthropic-AI',         'anthropic-ai'),
+    ('Google-Extended',      'google-extended'),
+    ('GoogleOther',          'googleother'),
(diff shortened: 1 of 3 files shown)


ManuShift:
## How to test

1. Import a log file with `--enable-bots` (old behavior, Custom Variables)
2. Import a log file with `--enable-bots=1` (new behavior, Custom Dimension 1)
3. Check in Matomo that bot visits show the classified bot name in the dimension

Example:
python import_logs.py --url=https://your-matomo.example.com --idsite=1 --enable-bots=1 access.log

ManuShift:
Friendly ping — would appreciate a review when you have time. Happy to address any feedback for using it in our productive system.

import_logs.py
```diff
@@ -844,8 +957,13 @@ def _create_parser(self):
         )
         parser.add_argument(
             '--enable-bots', dest='enable_bots',
-            action='store_true', default=False,
-            help="Track bots. All bot visits will have a Custom Variable set with name='Bot' and value='$Bot_user_agent_here$'"
+            nargs='?', const=True, default=None,
```
caddoo: This breaks compatibility for people, when importing multiple log files:

 Multiple log files:

  import_logs.py ... --enable-bots access1.log access2.log

  ```┌────────┬───────────────┬───────────────────────────────┬─────────────────────────────────────────────────────────────────┐
  │        │  enable_bots  │             file              │                             Outcome                             │
  ├────────┼───────────────┼───────────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Before │ True          │ ['access1.log','access2.log'] │ Imports both                                                    │
  ├────────┼───────────────┼───────────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │        │               │                               │ argparse accepts. int('access1.log') raises → caught → silently │
  │ After  │ 'access1.log' │ ['access2.log']               │  falls back to Custom Variables mode. Only access2.log gets     │
  │        │               │                               │ imported, with no warning that the first file was consumed by   │
  │        │               │                               │ the flag.                                                       │
  └────────┴───────────────┴───────────────────────────────┴─────────────────────────────────────────────────────────────────┘```

Can you maybe consider adding a new specific flag for what you are trying to do?

e.g. `--bot-custom-dimension N`

import_logs.py
```diff
@@ -2218,12 +2336,27 @@ def _get_hit_args(self, hit):
...
-        # handle custom variables before generating args dict
+        # handle bot tracking
         if config.options.enable_bots:
-            if hit.is_robot:
-                hit.add_visit_custom_var("Bot", hit.user_agent)
+            if config.options.enable_bots is True:
+                dimension_id = None
             else:
-                hit.add_visit_custom_var("Not-Bot", hit.user_agent)
+                try:
```
caddoo: Can we add some kind of arg type errors here for the dimension ID incase it's of an invalid type, just so it doesn't fail silently. Just a bit of extra defense

tests/test_main.py
```diff
@@ -1591,3 +1591,58 @@ def test_bz2_parsing():
...
+def test_get_bot_name_for_custom_dimension():
+    """Test bot name lookup from user agent string."""
+    assert import_logs._get_bot_name_for_custom_dimension('Mozilla/5.0 (compatible; GPTBot/1.0)') == 'GPTBot'
+    assert import_logs._get_bot_name_for_custom_dimension('Mozilla/5.0 (compatible; ClaudeBot/1.0)') == 'ClaudeBot'
+    assert import_logs._get_bot_name_for_custom_dimension('Mozilla/5.0 (compatible; Googlebot/2.1)') == 'Googlebot'
+    assert import_logs._get_bot_name_for_custom_dimension('Mozilla/5.0 (compatible; bingbot/2.0)') == 'Bingbot'
+    assert import_logs._get_bot_name_for_custom_dimension('Mozilla/5.0 (Windows NT 10.0) Chrome/120.0') == 'Other Bot'
+
+
+def test_enable_bots_custom_variable_fallback():
```
caddoo, line 1649: Hey, thanks for adding tests for this.

I want to flag though: I think these two tests (test_enable_bots_custom_variable_fallback and test_enable_bots_custom_dimension) are passing vacuously, so they aren't actually exercising the new code path yet.

When I instrumented them locally, Recorder.recorders ends up empty in both cases:

1. logs/ncsa_extended.log is a single 301 redirect, and the tests set enable_http_redirects = False, so the only line gets filtered out before reaching the recorder.
2. With nothing in Recorder.recorders, the for hit in Recorder.recorders: … break loop never iterates, so the assertions inside never run — the test just exits clean.
3. Even if a hit did land, the mock Recorder at test_main.py:205 only collects hits — it doesn't call _get_hit_args, which is where your new dimension code actually lives. So the branch under test wouldn't be hit either way.
  Two ways to fix this:
    - construct a Hit directly and call Recorder._get_hit_args(hit) on it, asserting on the returned args, or
     - swap in a log line containing a bot UA (something matching EXCLUDED_USER_AGENTS, e.g. a Googlebot/GPTBot line) and let it flow through naturally.

A couple of small things while you're in there:
  - Could you reset Recorder.recorders = [] at the top? The other tests in this file do that and it'll help avoid cross-test bleed.
  - The German comments (# kein Dimension-ID, # Dimension-ID als String) — could you translate those to English so they match the rest of the file?
  - test_get_bot_name_for_custom_dimension looks great as-is. 👍

caddoo:
@sgiehl tagging you in on this one for a couple reasons:

- PR sends bots=1 but no recMode, so AI-bot hits go to Visits, not log_bot_requests. Dashboard stays empty. Should imported logs route via recMode=2?
- PR hardcodes its own AI-bot list, duplicating device-detector/bots.yml. Will drift. Send raw UA and let core categorise instead?
- Unknown dimensionN is silently dropped in CustomDimensionsRequestProcessor. Want core to log/warn?
- CHANGELOG calls Custom Variables "deprecated in 5+" — is that actually formalised anywhere?

sgiehl:
@caddoo `bots=1` allows tracking any kind of bot, while we currently only filter for AI related bots, so this is in general still sort of a valid use case. 
However, managing a separate AI Bots list and storing their names in custom dimensions might not be something we should aim to do here. That's basically what AI Chatbot tracking in Matomo is already good for.
But I understand that there might a need to track all bots and have all of them in some sort of overview, which Matomo currently does not serve.
In general custom dimensions should be used in favor of custom variables, as they more performant in terms of archiving, so the general approach to change that is legit here...

ManuShift:
Hey, 

thanks for your Review ! 

What's been addressed:

1. Replaced nargs='?' on --enable-bots with a dedicated --bot-custom-dimension=DIMENSION_ID flag. Invalid dimension IDs are now rejected by argparse automatically (no silent fallback)
2. Translated German comments to English
3. Rewrote the tests so Hit objects are constructed directly and _get_hit_args() is actually called, meaning the dimension logic is now properly exercised

And yes, its official, that custom_variable is deprecated. See this:
https://matomo.org/faq/how-to/guide-to-using-custom-variables-deprecated/

[Re @sgiehl's comment](https://github.com/matomo-org/matomo-log-analytics/pull/394#issuecomment-4748849020): 
The goal here is exactly the use case you described. Tracking all bots and having them in a single overview, which Matomo currently doesn't serve out of the box. The AI Chatbot tracking plugin covers AI bots specifically, but not crawlers, SEO tools, monitoring bots etc. in one place. Furthermore, the API doesn't work well when importing 2–4 million requests in a daily batch and breaks with 500 errors.

Regarding the hardcoded bot list: the intention is precisely to have human-readable bot names in the dimension (e.g. GPTBot, Googlebot, TikTokSpider) rather than raw user agent strings, so the Matomo dashboard gives a meaningful breakdown out of the box. We're aware this list may drift from device-detector/bots.yml over time. Happy to discuss whether there's a better way to source the names, but dropping the classification entirely would defeat the purpose.

